Issue #31, opened 2015-12-08
Random V 2.0 task items


cartazio:
based up on the lovely GSOC work by  @nkartashov at https://github.com/nkartashov/SplitMix (and associated repos https://github.com/nkartashov/htestu for big crush/small crush statistical validation along and then https://github.com/nkartashov/prng-bench for performance comparison), @zaxtax and I are starting some planning work for Random Version 2.0
(and fyi to @idontgetoutmuch @gbaz et al)  

let it be known
1) based upon experimental data thus far, splitmix  (subject to certain caveats that are mostly around bad seed subsets that can be easily addressed by porting applicable machinery from the java sibling and possibly pestering guy steele as suggested  by some folks at oracle labs i spoke with this week), splitmix is the only rng aside from PCG-Random implemented in haskell that has a sequential `next` operation that passes big crush
2) splitmix is the only RNG we know of implemented in haskell that supports a  `split` operation that passes big crush when used as the sequencing operation 
(i need to do further auditing of how those crush tests were setup, but thats another sub item)

the following (approximate) task items, in part leveraging the excellent work by @nkartashov should form the foundation for prepping a v2 of random 
- [ ] have fast gen float / double siblings (theres fast ways to generate unit interval floats, and unbiased ways, we cant have both at the same time )
- [x] modern travis CI setup
- [ ] moving the c code back into haskell
- [x] criterion benchmarks to compare
- [ ] test suite to compare impls
- [ ] test suite to validate determination of stream given several seeds
- [ ] migrating the bad seed mitigation stuff from the Java sibling
- [ ] emailing GUY!!!! (about the splitmix improvement work he's apparently up to, i'll handle this one next weekend)
- [x] doc updates 
- [ ] considering UX changes
- [ ] RandomT?? 
  -  [ ] Should monadic bind do a split (a la quicheck )
- [ ] cleanup big crush stuff and have that as an ancillary library  (does @nkartashov want to own that? its his core work, and its an incredibly valuable tool, though he's busy with grad school and i dont want to impose that on him )

idontgetoutmuch:
This is great news.

Do we have the results of the tests written up somewhere?

I had a communication from Guy in April:

"My one recommendation is that you not standardize on a single algorithm.
Rather, make a SplittableRandom type class and provide multiple implementations."

Are we saying that PCG is somehow not as suitable as splitmix? I guess a write up of the tests would help answer this question.

cartazio:
[...]

PR #32, opened 2015-12-24
Getting rid of TAB characters

trskop:
This causes a lot of warnings while compiling, since -fwarn-tabs is
enabled by default.

### DIFF
--- a/Benchmark/BinSearch.hs
+++ b/Benchmark/BinSearch.hs
@@ -1,5 +1,5 @@
 
-{- 
+{-
  Binary search over benchmark input sizes.
 
  There are many good ways to measure the time it takes to perform a
@@ -16,7 +16,7 @@
  An alternative approach is to kill the computation after a certain
  amount of time and observe how much work it has completed.
  -}
-module BinSearch 
+module BinSearch
     (
       binSearch
     )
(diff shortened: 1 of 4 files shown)


cartazio:
There's already a pr for this.  But also the next release of random is
going to be a total rewrite so it won't have the issue :)

cartazio:
this work will be subsumed by #31

xkollar:
Well... but till then... would you consider at least releasing minor release? This patch looks good and it would remove quite some warnings people encounter on daily basis (which make it harder to notice significant warnings)... And from bootstraping cabal...

Please consider accepting this patch (at least for some branch for minor release...).

cartazio:
Okkkay.  I'll dig into that later today.

trskop:
There is also another possibility. Tell GHC to ignore that warning by passing `-fno-warn-tabs` option by using either `ghc-options` field in package Cabal file, or `OPTIONS_GHC` pragma.

cartazio:
I'll be getting back on the horse this wekeend.

Also in what context are these warnings creating real problems ?

Siprj:
I think that warnings really are big issue... I have to believe my build script to correctly skip warnings from "random" library and if something happens in build process I have to go through all these warnings manually.

I agree with @xkollar, this PR looks good. I don't see any problem that may come up with releasing minor version.

cartazio:
what build script is doing -WERROR  on dependencies warnings? thats a bug
in your build script / ci process

but yes, i'll get to this soon, have some other release engineering
underway this week

PR #36, opened 2016-05-08
extract constants k1 = 2147483562, k2 = 2147483398

kuk0:


lehins:
RNG implementation was switched to [splitmix](https://hackage.haskell.org/package/splitmix), these constants are no longer relevant.  See #62

PR #37, opened 2016-05-08
no default implemenation for genRange

kuk0:
i'm not sure this default is a good idea...
being explicit is safer and may prevent bugs


lehins:
This is has been superseded by #62 and is now the default implementation. Also `genRange` has been deprecated.

PR #38, opened 2016-05-08
genRange should be constant for a RandomGen - enforce this with types

kuk0:
the comment says genRange should _not_ examine its argument - well,
we can enforce this by types and instead of having genRange with a
RandomGen argument, we can create a phantom type GenRange g and
let genRange :: GenRange g (with no arguments)

### DIFF
--- a/Benchmark/Makefile
+++ b/Benchmark/Makefile
@@ -1,7 +1,6 @@
 
 
-#OPTS= -O2 -Wall -XCPP
-OPTS= -O2 -Wall -XCPP -Werror
+OPTS= -O2 -Wall -XCPP
 
 all: lib bench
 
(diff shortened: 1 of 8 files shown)


cartazio:
Just to be clear the v2 work isn't on master yet and will be a complete
rewrite of a lot of this stuff :)

Looks like the middle of my day will be consumed my family events but I'll
try to get that branch into shape and public later today

Zemyla:
Some of the code would be simpler if you had a function like:

```haskell
getRange :: (RandomGen g) => g -> GenRange g
getRange = const genRange
```

cartazio:
Yeah, there's definitely a whole bunch of design ideas that are worth
enumerating and comparing.

lehins:
`genRange` is now deprecated as of random-1.2. See #62

PR #40, opened 2016-10-04
Mark loop breaker as INLINABLE rather than INLINE

mpickering:
This prevents a lot of noisy warnings when building GHC.

### DIFF
--- a/Benchmark/Makefile
+++ b/Benchmark/Makefile
@@ -1,7 +1,6 @@
 
 
-#OPTS= -O2 -Wall -XCPP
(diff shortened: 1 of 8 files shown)


cartazio:
Will do a merge and bugfix release for this. Maybe today!

Been a bit buried but v2 work is nearing the top of the queue :)

cartazio:
Ill make sure this problem doesn’t happen in the 1.2 release.  Working on
it

lehins:
Thank you @mpickering This function is gone though and we now have a different implementation of floating point random number generation. See #62

PR #41, opened 2016-11-21
Fix broken link to Hellekalek

bgeron:


cartazio:
hrm... while DOI are useful links, i dont think its acceptable for us to have a URL that lands on a paywall, i'll think of some approach, hrmm

bgeron:
I've added another direct link to a uni-hosted PDF. I think we should keep the DOI because it is a permanent reference and URL links rot heavily, especially on university websites. What do you think?

bgeron:
Bump.

lehins:
We no longer link to this paper. Also the fixed link is now broken too ;) Although pdf is downloadable.

PR #48, opened 2018-08-03
Add NOINLINE pragma to theStdGen

### LINKED ISSUE #46: Add NOINLINE pragma for theStdGen
Top-level values initialized in IO with `unsafePerformIO` can be inlined, causing much awfulness.  There was a case in IRC earlier that appeared to be caused by inlining `theStdGen`.  Regardless of whether that was actually the cause, it should definitely be marked NOINLINE.

Proposal: add the pragma as such:

```
{-# NOINLINE theStdGen #-}
theStdGen :: IORef StdGen
theStdGen  = unsafePerformIO $ do
    rng <- mkStdRNG 0
    newIORef rng
```

treeowl:
We never want to inline `theStdGen`, as that could lead to
multiple generators, each starting from the initial value.

Fixes #46

### DIFF
--- a/System/Random.hs
+++ b/System/Random.hs
@@ -566,6 +566,9 @@ theStdGen :: IORef StdGen
 theStdGen  = unsafePerformIO $ do
    rng <- mkStdRNG 0
    newIORef rng
+-- We never want this to inline this, because we could end up with
+-- multiple generators, each starting from the initial value.
+{-# NOINLINE theStdGen #-}
 
 -- |Applies 'split' to the current global random generator,
 -- updates it with one of the results, and returns the other.


cartazio:
Stdgen isn’t in master. The next release won’t have it. And will not do the same magic global unsafe IO trick for starting sampling 

If you have an example where you can use magic inline primops to make a bug with this I’ll do a bug fix release. But master doesn’t any of that stuff with the current code or the new stuff.

treeowl:
It certainly is in master. What branch are you actually talking about?

cartazio:
It’s on my GitHub fork / mirror

cartazio:
hrm, it was already on this repo, in the WIP 1.2 branch, i've updated master  to reflect that

PR #49, opened 2018-09-01
Insert build status image inline

ron-wolf:


lehins:
We now have a proper CI on travis with coverage, badges and all other goodies. Thanks anyways ;)
See #62

Issue #59, opened 2020-03-10
read :: StdGen fails for strings longer than 6


joeyh:
ghci> read "abcdef" :: StdGen
    36217 1
    ghci> read "abcdefg" :: StdGen
    ** * Exception: Prelude.read: no parse

This is contrary to the documentation: "It guarantees to succeed on any string. "

random-1.1  
ghc-8.6.5

cartazio:
The documentation in read or in random. ?

I do strongly recommend using the pcg package for Rng engineering today.

The next major release of random will have pcg as one of the default
options.

I’m not sure what you’re trying to do with reading here?  It doesn’t look
like a deserialize to me ?

cartazio:
What’s the intent or goal you want to achieve here ? That can help inform
some of the planned new library work

andrewthad:
It's a [guarantee provided in the docs for random](https://github.com/haskell/random/blob/22a2a16bd62edd553b4f7f2e9eedc26cbf8850d8/src/System/Random.hs#L183):

> In addition, 'reads' may be used to map an arbitrary string (not necessarily one produced by 'show') onto a value of type 'StdGen'. In general, the 'Read' instance of 'StdGen' has the following properties:
>
> * It guarantees to succeed on any string.
> * It guarantees to consume only a finite portion of the string.
> * Different argument strings are likely to result in different results.

This is a weird thing to do with `reads`. Removing the broken `Read` instance for `StdGen` is a reasonable solution and deleting the associated documentation would be a reasonable fix.

cartazio:
Derp.  Yeah. Agreed on all fronts.  Been in a cavern on new random and
forgot about this cute little rough edge.

idontgetoutmuch:
I'd recommend using [splitmix](https://hackage.haskell.org/package/splitmix) as

  * It's the fastest
 *  Passes e.g. bigCrush so plenty good enough for non-cryptographic applications
 * Has a good `split` function
 * Is used by [QuickCheck](https://hackage.haskell.org/package/QuickCheck)

See here for the details on speed and quality  (which you can run yourself ):

* https://alexey.kuleshevi.ch/blog/2019/12/21/random-benchmarks/
* https://github.com/tweag/random-quality/tree/master/results

cartazio:
Pcg does better on big crush last I checked than splitmix. But either is
good.

curiousleo:
> Pcg does better on big crush last I checked than splitmix.

[...]

Issue #58, opened 2020-01-05
Why does random for Float and Double produce exactly 24 or 53 bits?


Zemyla:
It seems like we could exploit the fact that those types have more precision closer to 0. The algorithm would look like this:

```haskell
randomDouble :: RandomGen g => g -> (Double, g)
randomDouble = rr where
  b :: Word64
  b = bit 53
  mask = b - 1
  r = 1.0 / fromIntegral b

  rr g | g `seq` False = undefined
  rr g = case randomR (0, mask) g of
    (i, g') | testBit i 52 -> seq x (x, g') where
      x = r * fromIntegral i
    (0, g') -> go0 r 53 g'
    (i, g') -> let
      cs = countLeadingZeros i - 11
      in case randomR (0, bit cs - 1) g' of
        (k, g'') -> seq x (x, g'') where
          x = r * fromIntegral (unsafeShiftL i cs .|. k) / fromIntegral ((bit cs) :: Word64)

  go0 rc sh g | rc `seq` sh `seq` g `seq` False = undefined
  -- Stop before hitting denormals, because those are a pain.
  go0 rc sh g | sh >= 1022 = (0.0, g)
  go0 rc sh g = case randomR (0, mask) g of
    (i, g') | testBit i 52 -> seq x (x, g') where
      x = rc * fromIntegral i
    (0, g') -> go0 (rc * r) (sh + 53) g'
    (i, g')
      | sh + cs >= 1022 -> (0.0, g')
      | otherwise -> case randomR (0, bit cs - 1) g' of
          (k, g'') -> seq x (x, g'') where
            x = rc * fromIntegral (unsafeShiftL i cs .|. k) / fromIntegral ((bit cs) :: Word64)
      where
        cs = countLeadingZeros i - 11
```

And then randomRFloating could be defined to take advantage of the increased precision near 0:

```haskell
randomRFloating :: (Fractional a, Ord a, Random a, RandomGen g) => (a, a) -> g -> (a, g)
randomRFloating = rrf0 where
  rrf0 (l, h) g = case compare l h of
    LT -> rrf l h g
    EQ -> (l, g)
    GT -> rrf h l g

  rrf l h g | l `seq` h `seq` g `seq` False = undefined
  rrf l h g | l >= 0 = case random g of
    (coef, g') -> seq x (x, g') where
      x = 2.0 * (0.5 * l + coef * (0.5 * h - 0.5 * l))
  rrf l h g | h <= 0 = case random g of
[...]

Issue #55, opened 2019-08-28
System/Random.hs:43:1: warning: [-Wtabs]


p-alik:
`stack build` with [lts-13.26](https://www.stackage.org/lts-13.26) warns

```
random              >
random              > /tmp/stack5350/random-1.1/System/Random.hs:43:1: warning: [-Wtabs]
random              >     Tab character found here, and in 74 further locations.
random              >     Please use spaces instead.
random              >    |
random              > 43 |         (
random              >    | ^^^^^^^^
```

@nagaflokhu mentioned it also in [#50](https://github.com/haskell/random/issues/50#issue-383642492)
> There are also warnings about tab characters further up cabal's message.

https://github.com/haskell/random/blob/cfdfe6f09ad414fde5b855cc5f90207533413241/System/Random.hs#L42-L43

lehins:
All compile time warnings have been taken care of. See #62

Issue #53, opened 2019-04-01
incorrect distribution of randomR for floating-point numbers


shaobo-he:
Hello,

It appears that when the upper bound is very small, floating-point values generated are not correctly distributed.

For example, when the upper bound is `1.0e-45::Float` which is the smallest subnormal number of single precision floating-point representation, `randomR(0, (1.0e-45::Float))` does not produce `1.0e-45`. Furthermore, if the upper bound is `4.0e-45::Float`, `6.0e-45` is generated as follows,

```haskell
filter (>4.0e-45) $ take 10 $ randomRs (0,(4.0e-45::Float)) $ mkStdGen 0
[6.0e-45,6.0e-45]
```

cartazio:
Yup.  The current / usual naive approach to generating a unit interval
doesn’t actually work well.

1)
But: most / all unit interval algorithms are half open, they give the
interval [0,1) and then things get rescaled
, so you should never expect the 1 in your interval unless it’s from
rounding...


2) the over shoot is definitely looking like a bug, I’m not sure if it’s a
bug in our understanding of floating point near the cut off of normalized
floats one.  What’s the size of an ULP at that scale? I’d love some help
understanding what’s the cause of this bug.

Great job finding / reporting this stuff.  Thank you for taking the time.

shaobo-he:
I'm not sure how ULP is defined for subnormals but they are multiple of the minimum subnormal number. Let's call it `MSN`. `1.0e-45` is one `MSN` and `4.0e-45` is three `MSN`s. The number is generate using the following formula,
```haskell
let (coef,g') = random g in
                  (2.0 * (0.5*l + coef * (0.5*h - 0.5*l)), g')
```
When the lower bound is `0`, it becomes, `2.0*(coef*(0.5*h))`. And when `h` is 3 `MSN`s, `0.5*h` rounds to 2 `MSN`s. So when `coef` is greater than 0.75, `coef*(0.5*h)` rounds to 2 `MSN`s as well, making the entire expression evaluate to 4 `MSN`s, which is `4.0e-45`.

There are also other values that could trigger values being greater than upper bound. I used sbv (https://hackage.haskell.org/package/sbv) to find counterexamples. It works pretty well.

curiousleo:
In version 1.2.0 (released yesterday), we decided not to attempt to give stronger guarantees for floating point numbers. This decision was taken after a long discussion, see the summary here: https://github.com/idontgetoutmuch/random/issues/113#issuecomment-624041080.

[...]

PR #61, opened 2020-05-13
V1.2 proposal

idontgetoutmuch:
# Context

Following @lehins' [performance analysis][analysis] of Haskell pseudo-random number libraries and the [ensuing discussion][analysis-discussion], @lehins, @idontgetoutmuch and @curiousleo with help from @Shimuuar [set out to improve `random`][announcement] as both an interface for and implementation of a pseudo-random number generator for Haskell.

Our goals were to fix [#25][issue-25] (filed in 2015) and [#51][issue-51] (filed in 2018), see "Quality" and "Performance" below.

In the process of tackling these two issues, we addressed a number of other issues too (see "Other issues addressed" below) and added a monadic interface to the library so monadic pseudo-random number generators can be used interchangeably with `random`, see "API changes" below.

[This PR](this-pr) is the result of that effort. The changes are considerable. To signal this, we propose to release this as version 1.2 (the previous released version is 1.1, from 2014).

However, the API changes are generally backwards-compatible, see "Compatibility" below.

# Quality ([#25][issue-25])

We created an [environment][quality-repo] for running statistical pseudo-random number generator tests, tested `random` v1.1 and `splitmix` using dieharder, TestU01, PractRand and other test suites and [recorded the results][quality-results].

The results clearly show that the `split` operation in `random` v1.1 produces pseudo-random number generators which are correlated, corroborating [#25][issue-25]. The `split` operation in `splitmix` showed no weakness in our tests.

As a result, we replaced the pseudo-random number generator implementation in `random` by the one provided by `splitmix`.

# Performance ([#51][issue-51])

@lehins' [performance analysis][analysis] has the data for `random` v1.1. It is slow, and using faster pseudo-random number generators via `random` v1.1 makes them slow.

By switching to `splitmix` and improving the API, this PR speeds up pseudo-random number generation with `random` by one to three orders of magnitude, depending on the number type. See [Benchmarks][benchmarks] for details.

# API changes

The `Random` typeclass has conceptually been split into [`Uniform` and `UniformRange`][uniform-vs-uniformrange]. The `Random` typeclass is still included for backwards compatibility. `Uniform` is for types where it is possible to sample from the type's entire domain; `UniformRange` is for types where one can sample from a specified range.

A new module [`System.Random.Monad`][random-monad] allows monadic pseudo-random number generators like `mwc-random` to be used via the new interface provided by `random`, see the docs for an [example of how this works][mwc-example].

In addition, the module provides a convenient way to [run pure generators in monadic code][pure-gen].

# Changes left out

There were changes we considered and decided against including in this PR.

Some pseudo-random number generators are splittable, others are not. A good way of communicating this is to have a separate typeclass, `Splittable`, say, which only splittable generators implement. After long discussions (see [this issue][split-issue] and [this PR][split-pr]), we decided against adding `Splittable`: the interface changes would either have been backwards-incompatible or too complex. For now, `split` stays part of the `RandomGen` typeclass. The new documentation suggests that [`split` should call `error`][split-docs] if the generator is not splittable.

Due to floating point rounding, generating a floating point number in a range [can yield surprising results][fp-caveats]. There [are techniques][fp-issue] to generate floating point numbers in a range with actual guarantees, but they are more complex and likely slower than the naive methods, so we decided to postpone this particular issue.

Ranges on the real number line can be inclusive or exclusive in the lower and upper bound. We [considered][clusivity-issue] [API designs][clusivity-pr] that would allow users to communicate precisely what kind of range they wanted to generate. This is particularly relevant for floating point numbers. However, we found that such an API would make more sense in conjunction with an improved method for generating floating point numbers, so we [postponed][clusivity-postponed] this too.

# Compatibility

We strove to make changes backwards compatible where possible and desirable.

The following changes may break existing packages:

- import clashes, e.g. with the new functions `uniform` and `uniformR`
- `randomIO` and `randomRIO` where extracted outside of `Random` class as separate functions, which means some packages need to adjust how they are imported
- `StdGen` is no longer an instance of `Read`
- requires `base >= 4.10` (GHC-8.2)

In addition, `genRange` and `next` have been deprecated.

We have built all of Stackage against the code in this PR, and confirmed that no other build breakages occurred.

For more details, see [this comment][compatibility-comment] and [the "Compatibility" section in the docs][compatibility].

# Other issues addressed

This PR also addresses [#26][issue-26], [#44][issue-44], [#53][issue-53], [#55][issue-55], [#58][issue-58] and [#59][issue-59], see [Issues Addressed][issues-addressed] for details.

[analysis]: https://alexey.kuleshevi.ch/blog/2019/12/21/random-benchmarks/
[analysis-discussion]: https://www.reddit.com/r/haskell/comments/edr9n4/random_benchmarks/
[announcement]: https://mail.haskell.org/pipermail/libraries/2020-February/030261.html
[benchmarks]: https://github.com/idontgetoutmuch/random/blob/v1.2-proposal/CHANGELOG.md#benchmarks
[clusivity-issue]: https://github.com/idontgetoutmuch/random/issues/113
[clusivity-postponed]: https://github.com/idontgetoutmuch/random/issues/113#issuecomment-624041080
[clusivity-pr]: https://github.com/idontgetoutmuch/random/pull/104
[compatibility]: https://htmlpreview.github.io/?https://raw.githubusercontent.com/idontgetoutmuch/random/haddock-preview/doc/System-Random.html#g:6
[compatibility-comment]: https://github.com/haskell/random/pull/61#issuecomment-628173793
[fp-issue]: https://github.com/idontgetoutmuch/random/issues/105
[fp-caveats]: https://htmlpreview.github.io/?https://raw.githubusercontent.com/idontgetoutmuch/random/haddock-preview/doc/System-Random-Monad.html#g:14
[issue-25]: https://github.com/haskell/random/issues/25
[issue-26]: https://github.com/haskell/random/issues/26
[issue-44]: https://github.com/haskell/random/issues/44
[issue-51]: https://github.com/haskell/random/issues/51
[issue-53]: https://github.com/haskell/random/issues/53
[issue-55]: https://github.com/haskell/random/issues/55
[issue-58]: https://github.com/haskell/random/issues/58
[issue-59]: https://github.com/haskell/random/issues/59
[issues-addressed]: https://github.com/idontgetoutmuch/random/blob/v1.2-proposal/CHANGELOG.md#issues-addressed
[mwc-example]: https://htmlpreview.github.io/?https://raw.githubusercontent.com/idontgetoutmuch/random/haddock-preview/doc/System-Random-Monad.html#g:13
[pure-gen]: https://htmlpreview.github.io/?https://raw.githubusercontent.com/idontgetoutmuch/random/haddock-preview/doc/System-Random-Monad.html#g:5
[quality-repo]: https://github.com/tweag/random-quality
[quality-results]: https://github.com/tweag/random-quality/tree/master/results
[random-monad]: https://htmlpreview.github.io/?https://raw.githubusercontent.com/idontgetoutmuch/random/haddock-preview/doc/System-Random-Monad.html
[split-docs]: https://htmlpreview.github.io/?https://raw.githubusercontent.com/idontgetoutmuch/random/haddock-preview/doc/System-Random.html#v:split
[split-issue]: https://github.com/idontgetoutmuch/random/issues/7
[split-pr]: https://github.com/idontgetoutmuch/random/pull/9
[this-pr]: https://github.com/haskell/random/pull/61
[uniform-vs-uniformrange]: https://htmlpreview.github.io/?https://raw.githubusercontent.com/idontgetoutmuch/random/haddock-preview/doc/System-Random-Monad.html#g:10


cartazio:
Thx for pushing an update. I’ll take a look as time permits.  It’s a weird spring for everyone

lehins:
Hey @cartazio, I understand that you are pretty busy and have a lot on your plate. If you don't have time to deal with maintaining `random` would you like me to take over? I'll be happy to keep it in top notch shape. Considering that it is no longer a boot library and does not affect GHC it shouldn't be hard decision to make. What do you say?

cartazio:
@lehins actually the root bottleneck has been anxiety about how to bridge the gap between where things are today, and the set of breaking changes needed to give everyone something amazing.

I think i have a path to that now. 

my homework is to read through this and compare against other needs.

edit: i started reading this, and its a great start! theres some factorings I'd want to shift around but this code is great.

meta remark: i sometimes have whats called "generalized anxiety disorder", so i sometime get super stressed out over things that might not seem sensible unless you have a long conversation :)

lehins:
@cartazio I totally understand. I just want to make sure we get this initiative moving, I would hate it to drag along for another couple years, especially considering that this PR contains some solid code, which was already tested and benchmarked against many libraries.

>  gap between where things are today, and the set of breaking changes needed to give everyone something amazing.

If you look at the changelog there is some really amazing stuff in there and the set of breaking changes to get it is really minimal. The things we noticed when compiling against Stackage:

* Import clashes for new functions: `uniform` and `uniformR`
* `randomIO` and `randomRIO` where extracted outside of `Random` class as separate functions, which means some packages need to adjust how they are imported
* There is no more `Read` instance for `StdGen`, which breaks a few packages that rely on that. I'd rather them not compile instead of break at runtime because `read` function works differently. Also `Read` for `StdGen` was proven problematic in the past, so I think it would be best not to have it at all.
* Lower bound was raised to `base >= 4.10` (GHC-8.2)
* Deprecation pragma on `genRange` and `next`

This is really the full list of breaking changes that we could find, which is pretty standard stuff for a major version upgrade: break some imports, remove an instance and add a deprecation pragma. We'll post later the full table of packages that did break and which ones we already have fixes for.

There are some packages that needs to be adjusted in order to get all the performance benefits of the improvements in this PR. The packages I am talking about are the ones that implement instances for `RandomGen` (`QuickCheck`, `tf-random` ...) and we have all of these changes sitting ready in the corresponding forks waiting for this PR to be released.

> theres some factorings I'd want to shift around but this code is great.

Glad you already like the code. With regards to "I'd want to shift around": I know how you like "surgical" chopping up of PRs and I'd like to ask you if we can communicate any such changes through a separate PR (either into the https://github.com/idontgetoutmuch/random fork or in some other branch in this repo). Reason why I am asking is not because I am afraid you'll mess it up or something, but because we had some extensive discussions about all the choices that were made in this PR and we might have some arguments against some piece of code being changed. We had a whole collection of different attempts at the API, implementation and the overall structure, all of which can be found on github, but I'll be happy to point you to any of those attempts and discussion.

curiousleo:
We've put a lot of effort into documentation. **[Here is a rendered preview.](https://htmlpreview.github.io/?https://raw.githubusercontent.com/idontgetoutmuch/random/haddock-preview/doc/index.html)**

NorfairKing:
@cartazio +1 from me. I've seen this work get me 3x - 10X increases in speeds of my property test suites.

lehins:
Hey @cartazio Hope your weekend is going well. Did you get a chance to look at this PR. Do you have any questions or concerns?

cartazio:
i did a first read, looks good. i'll be putting together a bunch of stuff
that does the changes relative to a squashed version of the current PR and
we can talk about them.

cartazio:
@NorfairKing  what  property test stuff in particular?

lehins:
@cartazio I was gonna recommend against squashing, because then all links to discussions on PRs on github will be lost. What would be the benefit of squashing all commits together?

lehins:
Nevermind, looks like @curiousleo got rid of all merge commits so that history is already discarded. I don't really agree with this, since I think it is pretty important to have a connection to the pull request from commit history, but if you want to squash all history anyways then it doesn't matter.

NorfairKing:
> @NorfairKing what property test stuff in particular?

This change will improve the performance of `QuickCheck`s `Gen`s.

curiousleo:
> Nevermind, looks like @curiousleo got rid of all merge commits so that history is already discarded. I don't really agree with this, since I think it is pretty important to have a connection to the pull request from commit history, but if you want to squash all history anyways then it doesn't matter.

That sucks! Yes, I linearised the commit history by getting rid of all the merge commits. All other commits were preserved (which was my focus when I did it), but of course it's the merge commits that link the work to the PRs!

I am not knowledgeable enough about Git to (1) rewrite the history of a complex branch like `interface-to-performance` such that it appears to branch off `haskell/master` even though it initially didn't while (2) preserving merge commits. `v1.2-proposal` is the result of doing (1) without (2).

lehins:
The real problem here is that it doesn't make sense to branch of off `master` in the state it is now since the first commit in this PR replaces everything with random-1.1 

 @cartazio Do you think you could reset master to https://github.com/haskell/random/tree/v1.1 which is what currently on Hackage? Having master in such an unfinished state prevents anyone from contributing, including us.

cartazio:
A squash per pr with girhub url and archive dot org / way back url backup
or an accumulated squash ?

After 1.2 I want to move the bad compat bits into an old-random package for
Historical / legacy support

Along side this, the current random repo has for historical accident
reasons like the entire pre git commit history for all of base when it was
managed as a separate repo. I’d like to have that become  the base for old
random. And I guess new random could be. Asked on 1.1 source as intial
commit

I’ll see about doing the repo surgery needed, though ideally if you can
have a branch with the sequence of pr squash’s if you’re keen on that
sorted out , then I’ll do the surgery after that in terms of reapplying the
commit diffs to a new base

lehins:
> A squash per pr with girhub url and archive dot org / way back url backup
or an accumulated squash ?

^ @cartazio I am not sure I understand what you are trying to say here.

> After 1.2 I want to move the bad compat bits into an old-random package for
Historical / legacy support
> Along side this, the current random repo has for historical accident
reasons like the entire pre git commit history for all of base when it was
managed as a separate repo. I’d like to have that become  the base for old
random. And I guess new random could be. Asked on 1.1 source as intial
commit

^ Is there any benefit of doing this? This history that goes back to the origin of time and includes chunks of base doesn't hurt anyone. Splitting into two separate repositories will only bring complexity and will result questions from people that are unaware "What is this old-random thing?"

> I’ll see about doing the repo surgery needed, though ideally if you can
have a branch with the sequence of pr squash’s if you’re keen on that
sorted out , then I’ll do the surgery after that in terms of reapplying the
commit diffs to a new base

^ @curiousleo just rewrote the commit history of this PR specifically so we get all the merge commits as well, which give us a way to go back to each individual PR that contains all the relevant discussions. 

@cartazio I would be eternally grateful if you did not do this:
> then I’ll do the surgery after that in terms of reapplying the commit diffs to a new base

I think the most important part are the actual changes introduced in this PR and a proper review and feedback from you would be a lot more valuable than an attempt to chop up the history. If you want to suggest any changes to this PR, please I am all ears. Just comment on a PR like it happens in any usual PR review. Once we are all satisfied with the PR let's hit merge button and not destroy all the  history by "surgery"

I am also not following this one: 
> ideally if you can have a branch with the sequence of pr squash’s if you’re keen on that
sorted out

curiousleo:
@lehins wrote:
> @cartazio Do you think you could reset master to https://github.com/haskell/random/tree/v1.1 which is what currently on Hackage? Having master in such an unfinished state prevents anyone from contributing, including us.

@cartazio wrote:
> I’ll see about doing the repo surgery needed

---

@cartazio perhaps to clarify:
- the `master` branch is currently not in a state that can be used as the basis for a contribution
- the `v1.1` tag is still the best basis for contributions, including the one in this PR
- PRs can only be created against branches, not against tags, and there is no branch with the state of `v1.1`

To resolve this, I suggest a variant of the following to reset `master` to `v1.1`:

```
git checkout -b cartazio-master master
git checkout master
git reset --hard v1.1
git push -f origin master
```

No surgery required - this is just renaming things.

cartazio:
No.  :)

cartazio:
To be clear, I’m willing to do the for review purposes , but my end side
goal about long term git repos I already mentioned still stands :)

curiousleo:
@cartazio wrote:
> No. :)

Do you take issue with my premises or my conclusion? I'm happy to walk you through [my thinking](https://github.com/haskell/random/pull/61#issuecomment-630807743) here:

> * the `master` branch is currently not in a state that can be used as the basis for a contribution

The master branch [no longer has any tests](https://github.com/haskell/random/commit/bc34220740c467f0154d3da0eefc2b3a2db6ae96#diff-1615a42fbca6eaf32c4ce1c1fc9eaf1a) or [benchmarks](https://github.com/haskell/random/commit/bc34220740c467f0154d3da0eefc2b3a2db6ae96#diff-125af3b4ecf7a0b9c39cc2fddf2ce081) and has plenty of unfinished work in it, e.g. https://github.com/haskell/random/blob/615434ec425322ff2cab7d33e1c6211a91691972/src/Data/Distribution/FloatingInterval.hs#L44-L47

Pushing work in progress is _absolutely fine_, of course! But pushing unfinished work to `master` makes it difficult or impossible for others to contribute.

> * the `v1.1` tag is still the best basis for contributions, including the one in this PR

That's just a corollary from the previous point: `master` is not a good basis for contribution. The last released version was 1.1, and there is a tag for it, namely `v1.1`. So this is a tested and released version of the code, which makes it a good basis for contributing.

> * PRs can only be created against branches, not against tags, and there is no branch with the state of `v1.1`

That's just how GitHub works: you can only select an existing branch as a target for a PR.

---

My conclusion is that in order to allow contributions again, the best course of action would be to reset `master` to `v1.1`. I think this follows quite naturally.

cartazio:
i put a copy of this pr on wip/cartazio/v1.2-proposal-starter so i can have  a stable snapshot to work with :)

cartazio:
i'm looking at the commits now, and its quite literally based on top of master for some reason right now! :)

lehins:
Without basing it on top of `master` it was impossible to submit this PR.
> i'm looking at the commits now, and its quite literally based on top of master for some reason right now!

cartazio:
created a `release/v1.1-base` branch to reseat this against

cartazio:
One api detail I’m poking at is whether the end user api can be weakened to to applicative constraints via some backflips.

lehins:
>  end user api can be weakened to to applicative constraints

It can't be done for many reasons

* in order to generate a random value you need a generator that was created by generating previous value. Which is essentially a definition of bind. Splitting could possibly be abused to work around that, but few gens are splittable and that would have bad performance
* Same logic applies to stateful, but those have an actual state token passed around, which requires not only a monad, but PrimMonad
* generation of random ranges requires rejection, which means you need to be able to inspect the generated value and possibly perform another generation, which is prohibited in applicative

Most importantly, what is there to gain by relaxing to Applicative vs Monad, even if it was possible. Nothing useful!

cartazio:
Surprisingly the answer seems to be that it’s doable. :)

cartazio:
I’ll share in a little bit

cartazio:
Playing around a bit, it looks like rejection sampling is expressible via
Alternative mixed with Applicative.  Which also begets a very different
flavor for some api stuff.

idontgetoutmuch:
> Playing around a bit, it looks like rejection sampling is expressible via Alternative mixed with Applicative. Which also begets a very different flavor for some api stuff.

Won't this break lots of things? We've been very careful to minimize breakage in 1.2. Maybe this could form part of 2.0 (or even 3.0)?

I am also a bit sceptical about one one would do conditional sampling with only applicative. But we should also remember this library is here to provide base randomness on which e.g. distributions can built.

cartazio:
Indeed. This is making me think that perhaps for 1.2 we should table the
monadic apis generally?

I definitely admit this morning observation has me a teeny bit confused,
since it’s definitely the case that monadic vs monadic plus alternative
styles are somewhat different.

cartazio:
Or maybe there’s a way to support both.

cartazio:
I’m going to think about this more and cook up some concrete examples to
share


Also on the short bytestring front, I’d be more comfortable with figuring
out how to provide the write structure via primitive rather than
bytestring.  But I’ll work through that myself and share for feedback

NorfairKing:
@cartazio I'd like to get this merged rather sooner than later. It's the biggest performance improvement  in property testing to date. I frankly couldn't care less about an Applicative/Alternative-based APi.

cartazio:
@NorfairKing this is a code review discussion. Please either facilitate or refrain. Drive by comments do not help.

lehins:
>Also on the short bytestring front, I’d be more comfortable with figuring
out how to provide the write structure via primitive rather than
bytestring.  But I’ll work through that myself and share for feedback

@cartazio This is exactly why I asked for you to communicate back, so thank you. We already went through this process and in fact it was one of the first things I implemented was generation of `ByteArray` with `primitive`: https://github.com/idontgetoutmuch/random/pull/10/files 

But after giving it much thought and consideration we all agreed that it is better to have a separate random library that works with primitive, rather than making primitive a dependency of random, because

* Generating `ByteString` and `ShortByteString` is a lot more useful than `ByteArray`s and if need be `ShortByteStrings` can be easily converted to `ByteArray` in O(1) time
* There are a lot more packages that depend on random than there are ones that depend on `primitive`, we don't want to force all of those packages to suddenly pull `primitive` in for no huge benefit
* If ghc will need `random` again, they will have to make `primitive` boot library as well, which is certainly not a good idea.

Here is a PR and more of reasoning why depending on primitive is not a godd ide: https://github.com/idontgetoutmuch/random/pull/80#issuecomment-611086264

lehins:
So, as I said, impossible. You can't do rejection by using `Alternative` instance of `IO` monad, which is one of the most common places monadic interface will be use. `ST` does not have `Alternative` instance, but we need to be able generate random values in `ST`
> rejection sampling is expressible via Alternative mixed with Applicative

Remember, we want to make it possible to use `MoandRandom` with `PrimMonad` and `MonadState`, because otherwise this whole interface is useless. 

@cartazio Let's stay practical and focus on the review rather than new cool features that we could potentially implement. We've spent 3 moths on this and already got something solid, let's not add other cool features that might or might not be better. If we do that we will end up waiting another 6 years for a release. I know how you like to go on a tangent of great ideas, let's deal with what we have now and postpone the great potential features for future releases.

curiousleo:
Note to everyone watching this PR: the description has been updated to provide more context for the changes.

cartazio:
I’ll merge it in later today and prep a release with this work as the bassis.

cartazio:
Marking as closed since the merging won’t work correctly via gui,

lehins:
@cartazio You can give us a base branch where you'd like us to submit this PR to, this way we can adjust the history and still merge this PR via GUI
> Marking as closed since the merging won’t work correctly via gui,

cartazio:
wip/carter/1.1-base-forv1.2-collab  which should match the v1.1 tag

cartazio:
Plz holler when it’s done. 

Also if the commits could be squashed down a tad that would be great.

cartazio:
I should add , having spoken with a few folks and based on my own knowledge, Random was only in the boot set because it was originally part of base

lehins:
@cartazio its done

lehins:
> Also if the commits could be squashed down a tad that would be great.

I don't know if too much squashing is a good thing. Most of the history in this PR is fairly clean and commits are self-contained and buildable. Plus as I said it will be good to keep merge commits around to keep the link to original PRs with reviews and discussions.

> I should add , having spoken with a few folks and based on my own knowledge, Random was only in the boot set because it was originally part of base

Originally maybe, but according to [this table](https://gitlab.haskell.org/ghc/ghc/-/wikis/commentary/libraries/version-history) `random` was included with ghc-8.4.2, and some versions before it but not all of them. Which can't be explained by  the fact that random was part of base years ago

cartazio:
That may very well be, but pritmive is pretty stable so that isn’t a
concern for anyone In ghc land so it shouldn’t be yours either

lehins:
>That may very well be, but pritmive is pretty stable so that isn’t a
concern for anyone In ghc land so it shouldn’t be yours either

@cartazio This is not an argument in favor of using `primitive` for `random`

There are a lot of stable libraries in Haskell, it doesn't mean that random should depend on them. Can you provide reasons why `primitive` should be used with `random`? What are the benefits? Have you considered all the drawbacks?

I gave you a summary https://github.com/haskell/random/pull/61#issuecomment-631143490 of all reasons why we shouldn't use primitive, can  you respond to each individual one instead of telling me what people in ghc world think of `primitive`:

> * Generating `ByteString` and `ShortByteString` is a lot more useful than `ByteArray`s and if need be `ShortByteStrings` can be easily converted to `ByteArray` in O(1) time
> * There are a lot more packages that depend on random than there are ones that depend on `primitive`, we don't want to force all of those packages to suddenly pull `primitive` in for no huge benefit
> * If ghc will need `random` again, they will have to make `primitive` boot library as well, which is certainly not a good idea.

> Here is a PR and more of reasoning why depending on primitive is not a god idea: https://github.com/idontgetoutmuch/random/pull/80#issuecomment-611086264


FYI, it is up to me to decide what I should be concerned about.

cartazio:
cool :) 

ok, i'll include this changeset as a base, and do some tweaking to have a release candidate for the core important improvments that we all agree on! :)

lehins:
@cartazio I though you are going to merge this PR? Why did you close it again?

cartazio:
shit, id din't see you repointed it, my bad :)

cartazio:
i totally respect your persectives here (we're both right overall), and you're right about scope. the improvements for the APIS will all land, and i'm thinking for now we table the monadic apis from the actual release. 

even if i'm sometimes a tad "grrr", that work is outstanding, and shifting out the default RNG and not letting other things block it is on me.

i will do some other RNG shuffling, but i think you'll be happy overall (though perhaps rightly frustrated with me :) ).

curiousleo:
Hi @cartazio, re squashing: I also think that the commit history of this PR is valuable. A lot of thinking went into the decisions that were made, and the commit history points to the PRs where the relevant discussions took place.

There are definitely commits in the history that by themselves don't add much value, so there is an opportunity for careful squashing.

I volunteer to squash the history of this branch into one commit per PR. But only if the entire history does _not_ get squashed into one big commit in the end. Can you confirm?

curiousleo:
https://github.com/haskell/random/pull/61#issuecomment-631548500 ah I guess that's a moot point now :)

cartazio:
@curiousleo i'm going to persist this entire diff in a design-artifact/v1.2-proposal, so dont worry :) 

i'm also going to probably see about re-applpying the diff set ontop of a compacted/filtered history with v1.1 as the historical base for the new-new repo (and i"ll see about making sure all urls also point to internet archive backups if needed)

i'm actually frustrated with myself for not seeing the applicative+alternative design path YEARS ago, but c'est lie-vie

I hope that in aggregate you all find yourself more happy than annoyed with how i use this work, but i'm in a good headspace to engage on this stuff and focus (despite the plague times), and zooming out, we ALL care about good randomness :) 

i'm thinking for near term i'll have a child / satellite lib for workshopping out the fun fancy stuff

cartazio:
design_artifacts/theLeo_and_Lehins_etal_v1.2_proposal is the stable in repo snapshot for now

curiousleo:
> @curiousleo i'm going to persist this entire diff in a design-artifact/v1.2-proposal, so dont worry :)

That is appreciated, but beside the point: I think that a lot of the value in a good Git history lies in being able to answer the question "why is this code the way it is?" using `git blame`. This requires the `master` branch to have that history, since that is what people generally check out and run `git blame` on.

lehins:
>  i'm thinking for now we table the monadic apis from the actual release.

@cartazio monadic part of the API is critical to all of the work in this PR, cutting it out of the release will be a huge mistake

cartazio:
If we go that direction, I’d much rather use Brent yorgies monad random
design. I asked him many months ago if I could migrate those into

Plus making sure that that the base class has a liftSampling prim operation
with the type “g->(g,a)-> m a ”


That would work for both pure rng impls and runSt /primmonad compatible
implementations .


I’ll show y’all one that works with the mutable stuff you care about
instance wise.  And would support  both styles.

Have a look at Brent’s olde but good monad random type classes and please
tell me what you think!  There’s probably some opportunities for polish
there but I think it’s a good design

lehins:
@cartazio We have a concrete design that already works. Instead of saying what is possible could you please explain what you don't like about the current design and why do you think it doesn't work?

I am very well aware of all other approaches taken in all other libraries that provide alternative monadic interface. I know for a fact that none of them have the desired property of being suitable for both pure and stateful generators at the same time, including Brent's MonadRandom package. 

We've iterated on the design quite a bit and weighed all pros and cons. @Shimuuar who is the maintainer for the most popular satetful RNG library mwc-random has also participated in the design process and is onboard with this monadic interface.

> Plus making sure that that the base class has a liftSampling prim operation
with the type “g->(g,a)-> m a ”

Why do you need this extra function, this is just the `state :: (s -> (a, s)) -> m a` function from `MonadState`, so what's the point of reinventing stuff here. Current design already works with any monad that implements `MonadState`.

Current design is the only one that provides such great performance and usability across all available non-crypto RNGs in Haskell, it works with all transformers without any need to create crap load of instances, there is literally no more work needs to be done, so could you please shed the light for me why are you so reluctant about it?

cartazio:
To be concrete I mean https://hackage.haskell.org/package/MonadRandom-0.5.1.2/docs/Control-Monad-Random-Class.html

lehins:
I know, and I am saying that I know about it, I studied it amongst other approaches and it does not work with mwc-random and alike

idontgetoutmuch:
@cartazio you will no doubt be aware that many years ago it was proposed to make MWC compatible with `MonadRandom`.

https://github.com/byorgey/MonadRandom/issues/26

It's not clear from the discussion if this is feasible without changing `MonadRandom` in a breaking way.

I think you are looking at this the wrong way up. With this proposal, there is no need for `MonadRandom`.

cartazio:
... umm, @lehins  ... mwc doesnt have a monadic interface for threading the generator.

cartazio:
(i guess that lives in random-fu)

lehins:
Current PR has `MonadRandom` class that can be used with mwc-random, see the haddock in this PR it has examples on how it works

idontgetoutmuch:
I was just about to say the same thing :)

@cartazio see here:

https://htmlpreview.github.io/?https://raw.githubusercontent.com/idontgetoutmuch/random/haddock-preview/doc/System-Random-Monad.html#g:12

Shimuuar:
> ... umm, @lehins ... mwc doesnt have a monadic interface for threading the generator.

What is `PrimMonad m` that appears everywhere then :)? Instead of passing state explicitly it passes state token around and modifies generator in place. 

Although it's possible to have some vaguely `MonadRandom`ish interfaces for mwc-random for monads like `ReaderT (Gen (PrimState m)) m` it does require breaking changes.

cartazio:
Oooo.  I see what you mean. Brents stuff as is doesn’t work for you.  And
also isnt quite what I meant.

I’m gonna go play around and show you want I meant.  Might be a Friday or
next Monday (weekend fam trip).

cartazio:
Absolutely agree mwc needs to be sane to have a monad wrapper for.

curiousleo:
> Absolutely agree mwc needs to be sane to have a monad wrapper for.

With this PR, it does. Literal copy from the docs section "How to implement MonadRandom" (see https://htmlpreview.github.io/?https://raw.githubusercontent.com/idontgetoutmuch/random/haddock-preview/doc/System-Random-Monad.html#g:12), as @idontgetoutmuch pointed out:

---

Here is an example instance for the monadic pseudo-random number generator from the `mwc-random` package:

```haskell
instance (s ~ PrimState m, PrimMonad m) => MonadRandom MWC.Gen s m where
  type Frozen MWC.Gen = MWC.Seed
  thawGen = MWC.restore
  freezeGen = MWC.save
  uniformWord8 = MWC.uniform
  uniformWord16 = MWC.uniform
  uniformWord32 = MWC.uniform
  uniformWord64 = MWC.uniform
  uniformShortByteString n g = unsafeSTToPrim (genShortByteStringST n (MWC.uniform g))
```

cartazio:
It doesn’t need the state token in the monad random though.

lehins:
> It doesn’t need the state token in the monad random though.

@cartazio please elaborate, best with example.

If I am inferring correctly what you are saying, then it does not need `s` for `PrimMonad` becasue it works for all `s ~ PrimState m`. Loosing `s` though will prevent it from working in with things like gen stored in `IORef` and monad being `StateT`.  In short, It DOES need the state token in the monad random!

idontgetoutmuch:
Another data point: [random-fu](https://hackage.haskell.org/package/random-fu) is about x3 faster using the proposed new version:

random-fu-perf

PR #62, opened 2020-05-27
V1.2 proposal

lehins:
# Context

Following @lehins' [performance analysis][analysis] of Haskell pseudo-random number libraries and the [ensuing discussion][analysis-discussion], @lehins, @idontgetoutmuch and @curiousleo with help from @Shimuuar [set out to improve `random`][announcement] as both an interface for and implementation of a pseudo-random number generator for Haskell.

Our goals were to fix [#25][issue-25] (filed in 2015) and [#51][issue-51] (filed in 2018), see "Quality" and "Performance" below.

In the process of tackling these two issues, we addressed a number of other issues too (see "Other issues addressed" below) and added a monadic interface to the library so monadic pseudo-random number generators can be used interchangeably with `random`, see "API changes" below.

[This PR](this-pr) is the result of that effort. The changes are considerable. To signal this, we propose to release this as version 1.2 (the previous released version is 1.1, from 2014).

However, the API changes are generally backwards-compatible, see "Compatibility" below.

# Quality ([#25][issue-25])

We created an [environment][quality-repo] for running statistical pseudo-random number generator tests, tested `random` v1.1 and `splitmix` using dieharder, TestU01, PractRand and other test suites and [recorded the results][quality-results].

The results clearly show that the `split` operation in `random` v1.1 produces pseudo-random number generators which are correlated, corroborating [#25][issue-25]. The `split` operation in `splitmix` showed no weakness in our tests.

As a result, we replaced the pseudo-random number generator implementation in `random` by the one provided by `splitmix`.

# Performance ([#51][issue-51])

@lehins' [performance analysis][analysis] has the data for `random` v1.1. It is slow, and using faster pseudo-random number generators via `random` v1.1 makes them slow.

By switching to `splitmix` and improving the API, this PR speeds up pseudo-random number generation with `random` by one to three orders of magnitude, depending on the number type. See [Benchmarks][benchmarks] for details.

# API changes

## `StatefulGen`

The major API addition in this PR is the definition of a new class [`StatefulGen`][class-statefulgen]:

```haskell
-- | 'StatefulGen' is an interface to monadic pseudo-random number generators.
class Monad m => StatefulGen g m where
  {-# MINIMAL (uniformWord32|uniformWord64) #-}

  uniformWord32 :: g -> m Word32 -- default implementation in terms of uniformWord64
  uniformWord64 :: g -> m Word64 -- default implementation in terms of uniformWord32
  -- plus methods for other word sizes and for byte strings
[...]

PR #65, opened 2020-06-15
Generate .travis.yml via haskell-ci

curiousleo:
Cherry-picked from https://github.com/idontgetoutmuch/random/pull/148.


curiousleo:
I'm actually not sure how the coveralls integration works - @lehins, do we need to register the `haskell/random` project on coveralls.io or something like that?

lehins:
I am not quite sure, I've never setup coveralls for packages that aren't on my account. I do see `haskell/random` in my coveralls account, but that doesn't mean I have access to control it. 

So maybe @Bodigrim can help us with that, since he's part of haskell org.

What needs to happen is
* coveralls enabled for `haskell/random`
* token from coveralls needs to be added as env var to travis CI
* then `stack-hpc-coveralls` can upload reports.


What I recommend though. Let's merge the current work from #62  into master and then we can fix travis and coveralls, it will be easier to do that on the actual repository rather than from a fork.

curiousleo:
@lehins wrote:

> What needs to happen is
> 
>     * coveralls enabled for `haskell/random`
> 
>     * token from coveralls needs to be added as env var to travis CI
> 
>     * then `stack-hpc-coveralls` can upload reports.

Step 1 looks to be done: https://coveralls.io/github/haskell/random exists - thanks @Bodigrim! Do you have access to Travis for this repo too to add the token as an environment variable?

curiousleo:
Closing in favour of https://github.com/idontgetoutmuch/random/pull/148.

PR #67, opened 2020-06-24
Add missing newline in haddocks

Bodigrim:
Cf. http://hackage.haskell.org/package/random-1.2.0/docs/System-Random-Stateful.html#g:14 after "If a is -Infinity and b is Infinity..."


lehins:
Good catch.

PR #68, opened 2020-06-25
Infinities as floating point range endpoints

### LINKED ISSUE #54: randomR could produce NaNs when the upper bound is infinity
When the upper bound is infinity, NaNs could be produced. For example,

```haskell
filter (\x-> x/= (1/0)) $ randomRs ((0::Float), 1/0) $ mkStdGen 0
[NaN,NaN,NaN
```

Although it's unclear what should be produced when the upper bound is infinity, I don't think NaNs should be there anyways.

curiousleo:
Fixes #54. Brings the behaviour in line [with the docs](https://hackage.haskell.org/package/random-1.2.0/docs/System-Random-Stateful.html#g:14), which state that if exactly one endpoint is infinite, then `uniformRM` returns that endpoint.


src/System/Random/Internal.hs
```diff
@@ -806,6 +809,9 @@ uniformDoublePositive01M g = (+ d) <$> uniformDouble01M g
 instance UniformRange Float where
   uniformRM (l, h) g
     | l == h = return l
+    | isInfinite l && isInfinite h = return (0/0) -- NaN
```
curiousleo, line 812: This was previously the case too, tested here: https://github.com/haskell/random/blob/11464aa406a3387dc61feab84149feb9f671a5fb/src/System/Random/Stateful.hs#L608-L611

But we need this guard now so we can distinguish "both endpoints are unequal infinities" (this case) from "one of the endpoints is an infinity" (the next two cases)..

test/Spec.hs
```diff
@@ -126,13 +126,21 @@ integralSpec px =
...
      forall a.
-     (SC.Serial IO a, Typeable a, Num a, Ord a, Random a, UniformRange a, Show a)
+     (SC.Serial IO a, Typeable a, Num a, Ord a, Random a, UniformRange a, Read a, Show a)
   => Proxy a -> TestTree
 floatingSpec px =
   testGroup ("(" ++ showsType px ")")
   [ SC.testProperty "uniformR" $ seeded $ Range.uniformRangeWithin px
+  , testCase "r = +inf, x = 0" $ positiveInf @?= fst (uniformR (0, positiveInf) (ConstGen 0))
+  , testCase "r = +inf, x = 1" $ positiveInf @?= fst (uniformR (0, positiveInf) (ConstGen 1))
+  , testCase "l = -inf, x = 0" $ negativeInf @?= fst (uniformR (negativeInf, 0) (ConstGen 0))
```
curiousleo, line 136: This test fails without the changes to `uniformRM`.

curiousleo:
The benchmarks show an unacceptable decline in performance. I'll turn this back into a draft until I've figured this out.

Before (11464aa):
```
pure/random/Float                        mean 27.71 μs  ( +- 555.9 ns  )
pure/uniformR/unbounded/Float            mean 31.27 μs  ( +- 3.095 μs  )
```

Currently (ef2ee10):
```
pure/random/Float                        mean 15.57 ms  ( +- 2.282 ms  )
pure/uniformR/unbounded/Float            mean 15.89 ms  ( +- 2.239 ms  )
```

curiousleo:
64c52a5 makes `Float` perform as before, but `Double` takes twice as long as the baseline ...

Before (11464aa):
```
pure/random/Float                        mean 31.08 μs  ( +- 4.587 μs  )
pure/uniformR/unbounded/Float            mean 31.72 μs  ( +- 3.938 μs  )
pure/uniformR/floating/pure/uniformFloat01M mean 30.22 μs  ( +- 3.292 μs  )
pure/uniformR/floating/pure/uniformFloatPositive01M mean 28.35 μs  ( +- 1.074 μs  )

pure/random/Double                       mean 28.02 μs  ( +- 540.9 ns  )
pure/uniformR/unbounded/Double           mean 31.69 μs  ( +- 2.741 μs  )
pure/uniformR/floating/pure/uniformDouble01M mean 30.83 μs  ( +- 2.935 μs  )
pure/uniformR/floating/pure/uniformDoublePositive01M mean 28.66 μs  ( +- 1.471 μs  )

```

After (64c52a5):
```
pure/random/Float                        mean 27.79 μs  ( +- 661.0 ns  )
pure/uniformR/unbounded/Float            mean 30.00 μs  ( +- 3.122 μs  )
pure/uniformR/floating/pure/uniformFloat01M mean 27.83 μs  ( +- 515.2 ns  )
pure/uniformR/floating/pure/uniformFloatPositive01M mean 33.69 μs  ( +- 8.026 μs  )

pure/random/Double                       mean 56.83 μs  ( +- 1.486 μs  )
pure/uniformR/unbounded/Double           mean 31.34 μs  ( +- 3.889 μs  )
pure/uniformR/floating/pure/uniformDouble01M mean 32.32 μs  ( +- 4.288 μs  )
pure/uniformR/floating/pure/uniformDoublePositive01M mean 29.34 μs  ( +- 2.207 μs  )
```

curiousleo:
Performance issues fixed.

src/System/Random.hs
```diff
@@ -298,10 +298,10 @@ instance Random Char
 instance Random Bool
 instance Random Double where
   randomR r g = runStateGen g (uniformRM r)
-  random g = runStateGen g (uniformRM (0, 1))
+  random g = runStateGen g uniformDouble01M
```
curiousleo, line 301: Optimisation, should not change behaviour in any way.

Shimuuar:
> makes Float perform as before, but Double takes twice as long as the baseline ...

I've observed such behavior when I implemented `uniformDouble01M` changes such as adding/removing benchmark led to 2x performance jumps. It looks like GHC might run or skip some optimization

src/System/Random.hs
lehins: Looks like somehow there was a change in behavior, one doctest is failing, but I don't see why it would fail

curiousleo: This is because `uniformR (0, 1) == 1 - uniformFloat01M`. I have changed the test accordingly: https://github.com/haskell/random/pull/68/files#diff-911d03db4d77c6ca7a8a6a8865daa8f3L279-R279

```diff
- -- 0.5728354935654512
+ -- 0.42716450643454884
```

where 1 - 0.5728354935654512 = 0.42716450643454884.

curiousleo:
Ready for review.

test/Spec.hs
```diff
@@ -126,13 +126,21 @@ integralSpec px =
...
+  , testCase "l = -inf, x = 1" $ negativeInf @?= fst (uniformR (negativeInf, 0) (ConstGen 1))
   -- TODO: Add more tests
   ]
+  where
+    positiveInf, negativeInf :: a
+    positiveInf = read "Infinity"
```
idontgetoutmuch, line 142: I would write positiveInf = 1 / 0 then you don't need the Read constraint but I don't know if this is good practice.

Shimuuar: It is certainly much faster. Probably even constant-folded

lehins: Considering that it is in a test suite, I don't think performance matters much. Using `read` makes it a bit more obvious I think.

src/System/Random.hs
curiousleo: ... so this actually does change behaviour. Strictly speaking, this goes against the Reproducibility guarantee: https://hackage.haskell.org/package/random-1.2.0/docs/System-Random.html#g:8

We could
- (A) make the change anyway
- (B) make the change and release the next version as a major version
- (C) change this call so it preserves behaviour

If (C) is not slower than what is currently there, I'm in favour of (C). I'm going to try it out.

curiousleo: See https://github.com/haskell/random/pull/68/commits/2cbfe1bde3b38190976356a9bb7f6948dde306dd. The benchmarks don't show a slowdown.

lehins: I like the C option too. Thank you for adjusting this.

Shimuuar: > this actually does change behaviour. 

Infinity handling changes behavior too albeit subtly. `uniformRM (0,+∞) g >> uniformM g` will produce different results before and after this PR. Before Word32/64 was generated for uniformRM after generator state is not changed.

lehins: Good point. It might be good plan for the next version to be `random-1.3.0`. This would allow us to add other minor, but breaking changes, for example deprecation pragmas for the global `theStdGen` and related functions.

Bodigrim: I'm happy with (C) here, but do not see any downsides in (B).

Bodigrim:
I think this is ready to be merged. @Shimuuar @idontgetoutmuch if either of you has a strong opinion about changing `read "Infinity"`, please emphasize it.

Shimuuar:
Sorry, I forgot about this PR. Why NaN is returned when both endpoints are infinities? Following definition (in pseudocode) seems more sensible:
```haskell
uniformRM (+∞,+∞) _ = return +∞
uniformRM (-∞,-∞) _ = return -∞
uniformRM (-∞,+∞) _ = oneOf [-∞,+∞]
```

Also number of ifs on happy path (neither endpoint is ∞) worry me a bit.

Bodigrim:
> Why NaN is returned when both endpoints are infinities?
This has been discussed back in #54. I agree with @idontgetoutmuch that returning NaN is more reasonable than alternation between infinities.

Bodigrim:
From my point of view the issue in #54 is not that much about returning NaNs _per se_, but about returning NaNs inconsistently, only in a fraction of cases, so small, that it would easily pass a test suite of 100 random values. It is worth to change this behaviour a) to be consistent with the majority of cases, b) to match our own documentation: http://hackage.haskell.org/package/random-1.2.0/docs/System-Random-Stateful.html#g:14

On the other hand, `uniformRM (-∞,+∞)` returns `NaN` consistently and is clearly documented ibidem. I do not see much reason to change this behaviour.

Shimuuar:
I just found this behavior _mildly_ surprising which is fine for floating point. What isn't when it comes to floating point. 

BTW I came up with optimization which exploits quirks of floating point arithmetics:
```haskell
  | isInfinite l || isInfinite h = return $! h + l
```
If only one of `l`,`h` is infinite they'll absorb finite counterpart and (-∞) + (+∞) = NaN. It bring implementation from 4 ifs to only 2 ifs.

Otherwise LGTM

Bodigrim:
@curiousleo what do you think about @Shimuuar's suggestion? Also, could you please rebase?

curiousleo:
> @curiousleo what do you think about @Shimuuar's suggestion? Also, could you please rebase?

Thanks for the reminder. The suggestion looks neat! I'll come back to this PR as soon as free time permits.

Issue #26, opened 2015-04-23
Add Random instances for tuples


ndmitchell:
Looking at the interface, I see no reason tuples can't be instances of the Random class, which would make it more useful.

cartazio:
That sounds super reasonable. I'll talk about it with Dominic when we next
catchup.

I assume we probably don't need more Than size 6 or 10 tuples? :)

ndmitchell:
My particular use case was for 4 tuples - specifically, I wanted to call [`seedTFGen`](https://hackage.haskell.org/package/tf-random-0.5/docs/System-Random-TF-Gen.html#v:seedTFGen) with a randomly chosen seed. Agreed, 6 seems sufficient.

cartazio:
@ndmitchell this is a good feature, i'll make it happen, see what i can do...

how would you specify/what to talk about generating samples / intervals?

cartazio:
(i'm amidst revisiting / generalizing the sampler api, though that wont happen for the next release, i presume you want every tuple value, not an "interval/range"?

ndmitchell:
@cartazio I didn't really have any idea about what intervals should mean - but I guess between `(a,b)` and `(c,d)` should probably be equivalent to the first member in the range `a-c` and the second in `b-d`.

power-fungus:
What is the status on this issue? Is there a chance that the pull-request will be merged and an update is published?

cartazio:
Support for tuple sampling will be happening. Stay tuned

L-TChen:
Will this be happening soon? I am really looking forward to this feature. :-)

lehins:
There are now up 7-tuple instances for `Uniform` type class, so by using [`uniform`](https://hackage.haskell.org/package/random-1.2.0/docs/System-Random.html#v:uniform) it is already possible to get random tuples. I don't see a reason why we can't add such instances to `Random` as well. There was too many changes already in `random-1.2`, so we'll add those instances as well as a few other types in the next release.

Bodigrim:
`Random` for tuples requires implementing `randomR` to generate a random value in a range. Should we define `UniformRange` instances for tuples as well then?

Shimuuar:
[...]

Issue #21, opened 2014-10-04
Please provide instance for (Enum a, Bounded a) for Random


kvanbere:
It would be really nice to be able to do something like this:

```
> data Action = Code | Eat | Sleep deriving (Bounded, Enum)
> randomR :: IO Action
Eat
```

The implementation for `(Enum a, Bounded a) =>` can cheat a little to get a valid range (psuedo, for randomRIO since it's easier/shorter to show):

```
let lower = fromEnum (minBound :: a)
let upper = fromEnum (maxBound :: a)
return $ toEnum <$> randomRIO (lower, upper)
```

Note the usage of the underlying instance for `Int` to select a constructor.

It would have been magical if we could have just a `Enum a =>` instance, but I can't think of a way to do it.

vimuel:
This would be really nice to have.

donkeybonks:
Ah yes, another +1 on this .. it’s 3 year’s old, I had forgotten all about it .

donkeybonks:
Is PR accepted ?

cartazio:
I’m working on a new release.  Stay tuned.

Bodigrim:
I'm not convinced that sprinkling a couple of `fromEnum` / `toEnum` is that much of boilerplate. It certainly does not justify introducing overlapping instances. 

However, I would not particularly oppose a PR, introducing `uniformBoundedEnum :: (Enum a, Bounded a, StatefulGen g m) => g -> m a` or something similar.

Shimuuar:
I think we should introduce newtype wrapper which allows to derive Uniform/UniformRange instances using DerivingVia

```haskell
newtype UniformEnum a = UniformEnum a

instance Enum a => UniformRange (UnniformEnum a) where
  ...
instance (Bounded a, Enum a) => Uniform (UnniformEnum a) where
  ...

data RGB = R | G | B
  deriving (Enum.Bounded)
  deriving Uniform via UniformEnum RGB
```

chessai:
@Shimuuar that looks good to me!

Bodigrim:
Closing as done in #104. Feel free to reopen if there is a specific motivation to use this feature via old `Random` interface.

georgefst:
> ```hs
> newtype UniformEnum a = UniformEnum a
> 
> instance Enum a => UniformRange (UniformEnum a) where
>   ...
> instance (Bounded a, Enum a) => Uniform (UniformEnum a) where
>   ...
> 
> data RGB = R | G | B
[...]

PR #70, opened 2020-06-26
Uniform for Generic 

Bodigrim:
Cf. #21 and partially #26 

```haskell
> :set -XDeriveGeneric -XDeriveAnyClass
> import GHC.Generics (Generic)
> data Action = Code Bool | Eat (Maybe Bool) | Sleep deriving (Show, Generic, Uniform)
> gen <- newIOGenM (mkStdGen 42)
> uniformListM 10 gen :: IO [Action]
[Code False,Code False,Eat Nothing,Code True,Eat (Just False),Eat (Just True),Eat Nothing,Eat (Just False),Sleep,Code True]
```

### DIFF
--- a/src/System/Random/GFinite.hs
+++ b/src/System/Random/GFinite.hs
@@ -31,9 +31,16 @@ data Cardinality
   | Card  !Integer
   deriving (Eq, Ord, Show)
 
+-- | This is needed only as a superclass of 'Integral'.
 instance Enum Cardinality where
   toEnum = fromIntegral
   fromEnum = fromIntegral
+  succ = (+ 1)
+  pred = subtract 1
+  enumFrom x           = map fromInteger (enumFrom (toInteger x))
+  enumFromThen x y     = map fromInteger (enumFromThen (toInteger x) (toInteger y))
+  enumFromTo x y       = map fromInteger (enumFromTo (toInteger x) (toInteger y))
+  enumFromThenTo x y z = map fromInteger (enumFromThenTo (toInteger x) (toInteger y) (toInteger z))
 
 instance Num Cardinality where
   fromInteger 1 = Shift 0  -- ()
@@ -54,6 +61,7 @@ instance Num Cardinality where
   signum = Card . signum . toInteger
   negate = Card . negate . toInteger
 
+-- | This is needed only as a superclass of 'Integral'.
 instance Real Cardinality where
   toRational = fromIntegral
 
--- a/src/System/Random/GFinite.hs
+++ b/src/System/Random/GFinite.hs
@@ -171,7 +171,7 @@ instance (GFinite a, GFinite b) => GFinite (a :*: b) where
   {-# INLINE toGFinite #-}
 
   fromGFinite (q :*: r) =
-    toInteger (gcardinality (proxy# :: Proxy# a) * Card (fromGFinite q)) + fromGFinite r
+    toInteger (gcardinality (proxy# :: Proxy# b) * Card (fromGFinite q)) + fromGFinite r
   {-# INLINE fromGFinite #-}
 
 instance Finite Void
--- a/src/System/Random/GFinite.hs
+++ b/src/System/Random/GFinite.hs
@@ -1,0 +1,272 @@
+-- |
+-- Module      :  System.Random.GFinite
+-- Copyright   :  (c) Andrew Lelechenko 2020
+-- License     :  BSD-style (see the file LICENSE in the 'random' repository)
+-- Maintainer  :  libraries@haskell.org
+--
+
+{-# LANGUAGE DefaultSignatures    #-}
+{-# LANGUAGE FlexibleContexts     #-}
+{-# LANGUAGE LambdaCase           #-}
+{-# LANGUAGE MagicHash            #-}
+{-# LANGUAGE ScopedTypeVariables  #-}
+{-# LANGUAGE TypeOperators        #-}
+
+module System.Random.GFinite
+  ( Cardinality(..)
+  , Finite(..)
+  , GFinite(..)
+  ) where
+
+import Data.Bits
+import Data.Int
+import Data.Void
+import Data.Word
+import GHC.Exts (Proxy#, proxy#)
+import GHC.Generics
+
+-- | Cardinality of a set.
+data Cardinality
+  = Shift !Int -- ^ Shift n is equivalent to Card (bit n)
+  | Card  !Integer
+  deriving (Eq, Ord, Show)
+
+instance Enum Cardinality where
+  toEnum = fromIntegral
+  fromEnum = fromIntegral
+
+instance Num Cardinality where
+  fromInteger 1 = Shift 0  -- ()
+  fromInteger 2 = Shift 1  -- Bool
+  fromInteger n = Card n
+  {-# INLINE fromInteger #-}
+
+  x + y = fromInteger (toInteger x + toInteger y)
+  {-# INLINE (+) #-}
+
+  Shift x * Shift y = Shift (x + y)
+  Shift x * Card  y = Card (y `shiftL` x)
+  Card  x * Shift y = Card (x `shiftL` y)
+  Card  x * Card  y = Card (x * y)
+  {-# INLINE (*) #-}
+
+  abs    = Card . abs    . toInteger
+  signum = Card . signum . toInteger
+  negate = Card . negate . toInteger
+
+instance Real Cardinality where
+  toRational = fromIntegral
+
+instance Integral Cardinality where
+  toInteger = \case
+    Shift n -> bit n
+    Card  n -> n
+  {-# INLINE toInteger #-}
+
+  quotRem x' = \case
+    Shift n -> (Card (x `shiftR` n), Card (x .&. (bit n - 1)))
+    Card  n -> let (q, r) = x `quotRem` n in (Card q, Card r)
+    where
+      x = toInteger x'
+  {-# INLINE quotRem #-}
+
+-- | A type class for data with a finite number of inhabitants.
+-- This type class is used
+-- in default implementations of 'System.Random.Stateful.Uniform'
+-- and 'System.Random.Stateful.UniformRange'.
+--
+-- Users are not supposed to write instances of 'Finite' manually.
+-- There is a default implementation in terms of 'Generic' instead.
+--
+-- >>> :set -XDeriveGeneric -XDeriveAnyClass
+-- >>> import GHC.Generics (Generic)
+-- >>> data MyBool = MyTrue | MyFalse deriving (Generic, Finite)
+-- >>> data Action = Code MyBool | Eat (Maybe Bool) | Sleep deriving (Generic, Finite)
+--
+class Finite a where
+  cardinality :: Proxy# a -> Cardinality
+  toFinite :: Integer -> a
+  fromFinite :: a -> Integer
+
+  default cardinality :: (Generic a, GFinite (Rep a)) => Proxy# a -> Cardinality
+  cardinality _ = gcardinality (proxy# :: Proxy# (Rep a))
+
+  default toFinite :: (Generic a, GFinite (Rep a)) => Integer -> a
+  toFinite = to . toGFinite
+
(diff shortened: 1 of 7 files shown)


Shimuuar:
There's interesting question about efficiency for generation of pure product types. For large products (say `Foo Word64 Word64 Word64` it's obviously more efficient to generate each Word64 separately. Multiplication and division for long arithmetic is not terribly fast and superlinear in bit size. However for cases like `Bar D3 D3 D3` it could be more efficient to generate uniform number on range [0,27) and then do two divisions.

Bodigrim:
I'm not particularly concerned about performance here - `Generic` instances are expected to be less efficient than hand-written ones. It would be nice to pickup low-hanging fruits if any, but I do not want to complicate the design a lot.

Bodigrim:
Since `cardinality` is hidden from users, we can potentially make it not `Integer`, but `Either Int Integer`, where `Left 32` stands for 2^32 elements. And `GFinite (a :*: b)` could use this information, switching to shifts instead of divisions.

Shimuuar:
I disagree about performance. I certainly expect that performance of  generic instances would be similar to naively written ones it also should avoid pathological performance cases if it's possible. Otherwise we get "this is standard way to derive instances except don't use it this and that case" (it will be used anyway). And generic deriving is place where it's reasonable to spend optimization effort even if makes things more complicated. Cost is paid once but benefit is spread over all users.

Bodigrim:
Any ideas why `doctests` pass in all builds except GHC 8.10? https://travis-ci.org/github/haskell/random/jobs/702872773 CC @lehins

lehins:
> Any ideas why doctests pass in all builds except GHC 8.10?

@Bodigrim I am not sure, I've been seeing this with my own packages too, but I haven't investigated it much. Feels like something has changed with ghc-8.10 and Cabal combination which prevents doctests to function properly. I usually only care that doctests  are executed with some ghc version, because this way there is an already pretty good chance that examples in haddock are still good.

Really seems like an upstream bug and I try to stay away from digging into build tools too much and let someone else figure it out :)

We could for now disable doctests with CPP for ghc-8.10, they are already disabled for < ghc-8.2

If you feel like investigating it further though, please let me know what you find.

Bodigrim:
@lehins Yeah, I've seen this kind of arbitrary doctest failures with GHC 8.10 before, in my other projects.  Disabling for now.

@Shimuuar Well, one can certainly get more mileage generating made-to-measure instances via `TemplateHaskell`, but I'm certainly against this approach in `random`. However, I implemented the trick described above to replace long divisions/multiplications with shifts and sprinkled `{-# INLINE #-}` generously.

Shimuuar:
I thought a little more about it and come to conclusion that there's much better approach for product types. Current one defines uniform in terms of `Finite` which requires user to define `Finite` instance for user data types which isn't that terrible but it becomes impossible once you start working with real-valued data. Let take for example 

``` haskell
newtype Angle = Angle Double

instance Uniform Angle where
  uniformM g = Angle . ((2*pi)*) <$> uniformM g
```

It's perfectly valid uniform instance yet there's no sane `Finite` instance. Make it point on sphere and there's no instance at all.

I think better approach is to define generic `Uniform` for product in terms `Uniform` and fall back on `Finite` shenanigans for sum types. Something along the lines:

```haskell
class GUniform f where
  guniformM :: StatefulGen g m => g -> m (f p)

instance GUniform f => GUniform (M1 i c f) where
  guniformM = fmap M1 . guniformM

instance Uniform a => GUniform (K1 i a) where
  guniformM = fmap K1 . uniformM

instance GUniform U1 where
  guniformM _ = return U1

instance (GUniform f, GUniform g) => GUniform (f :*: g) where
  guniformM g = (:*:) <$> guniformM g <*> guniformM g

instance (...) => GUniform (f :+: g) where
  guniformM g = {- uses Finite -}
```

I think it's better since it avoids introducing `Finite` in common case of product types. So it appears only when dealing with sum types.

Bodigrim:
I'm not particularly opposed to your idea, but I have some reservations.

0. `Angle` is not quite a geometrical angle, because it includes both 0 and 2π. 
1. There are plenty of instances of `Uniform` for tuples, so if your data is a plain product you can whip up an instance of `Tuple Angle Angle` in no time. 
2. Under your proposal makes instances for small finitely-inhabited products worse: `Quadruple Bool Bool Bool Bool` would require 4 random numbers instead of 1 as currently.

Shimuuar:
0. It's depends on interpretation of data type. Sometime one want to consider 0 and 2π a same thing.BTW `2*pi`<2π. This is just a simplest example. It's very easy to invent more: just take any bounded set in R^n. 

1. I don't understand how tuples could help to define instance for entirely contrived `data Angle2 = Angle Angle Angle`. One could write `uniformM g = AngleA <$> uniformM g <*> uniformM g` but whole point of generics is to avoid such boilerplate. 

2. True. For types with small number of inhabitants it's more efficient to generate single random number and then divide. Working DerivingVia would've resolved this tension. However for single default implementation I think it's better to err on side of being more generic. We could provide `uniformViaFiniteM` hwen it could be used as optimzation

Bodigrim:
I actually thought a bit more and decided that it is better to follow `GUniform` avenue, because it allows users to inject their own instances of `Uniform`. Thanks, @Shimuuar!

Shimuuar:
> because it allows users to inject their own instances of Uniform

I had this idea in mind and couldn't explain it properly. Thanks for articulating it

src/System/Random/Internal.hs
```diff
@@ -499,9 +502,50 @@ class Uniform a where
...
+  guniformM = fmap M1 . guniformM
+
+instance Uniform a => GUniform (K1 i a) where
+  guniformM = fmap K1 . uniformM
+
+instance GUniform U1 where
+  guniformM _ = return U1
+
+instance (GUniform f, GUniform g) => GUniform (f :*: g) where
+  guniformM g = (:*:) <$> guniformM g <*> guniformM g
```
Bodigrim, line 535: This definition appears to be problematic: `Generic` representation fails to fuse because of monadic binds in the midst of `(:*:)`. Even for 
```haskell
{-# LANGUAGE DeriveGeneric, DeriveAnyClass #-}
data Tuple = Triple !Word !Word deriving (Generic, Uniform)
```
generated Core is unwieldy:
```haskell
-- RHS size: {terms: 62, types: 742, coercions: 210, joins: 0/5}
$fUniformTuple_$cuniformM
  :: forall g (m :: * -> *). StatefulGen g m => g -> m Tuple
```

`GFinite`-based version avoids this shortcoming by executing effectful computation exactly ones.

Bodigrim: When we derive `Uniform` via `GFinite`, cardinality is usually statically known, so we immediately generate a random `m Integer`, then create a generic representation `Integer -> Rep a` and convert it to a concrete data `Rep a -> a`. Since both latter steps are pure, GHC is happy to inline everything and avoid actual construction of `Rep a` altogether.

By using `GUniform` we construct a random `m (Rep a)`, then convert it to a concrete representation `Rep a -> a`. Because of a polymorphic monad `m`, GHC cannot really inline stuff through monadic binds and we end up constructing `Rep a` in full. 

I guess that in client's code, which works in a fixed monomorphic `m`, the definition of `uniformM` would inline specialized and actually avoid allocating `Rep a`. But this is a fragile assumption, so I decided to export `uniformViaFiniteM` to be on the safe side.

Shimuuar: @Bodigrim I made a quick test with converting to continuations. General idea is to work in `ContT m` instead of `m`

```haskell
class GUniform f where
  guniformM :: StatefulGen g m => g -> ContT r m (f a)
```

Since `ContT` monad instance doesn't use bind of underlying monad GHC is free to inline and otherwise specialize. And it's able to remove all traces of Generic from generated instance

```
from: -- RHS size: {terms: 61, types: 589, coercions: 246, joins: 0/8}
to:   -- RHS size: {terms: 17, types: 76, coercions: 0, joins: 0/0}
+     -- RHS size: {terms: 30, types: 51, coercions: 0, joins: 0/1}

```

Bodigrim: Awesome! What is `+`?

Shimuuar: GHC split definition in two functions. That's count for both

src/System/Random/GFinite.hs
```diff
@@ -0,0 +1,272 @@
...
+import GHC.Exts (Proxy#, proxy#)
+import GHC.Generics
+
+-- | Cardinality of a set.
+data Cardinality
+  = Shift !Int -- ^ Shift n is equivalent to Card (bit n)
+  | Card  !Integer
+  deriving (Eq, Ord, Show)
+
+instance Enum Cardinality where
```
Shimuuar, line 34: to/fromEnum go through Int so they're subject to overflow. Thus `succ (2^128)` won't be `2^128+1`. I think all methods should be either implemented using Integer everywhere or it should be dropped altogether

src/System/Random/Internal.hs
```diff
@@ -499,9 +506,73 @@ class Uniform a where
...
+  -- >>> uniformListM 10 gen :: IO [Action]
+  -- [Code MyTrue,Code MyTrue,Eat Nothing,Code MyFalse,Eat (Just False),Eat (Just True),Eat Nothing,Eat (Just False),Sleep,Code MyFalse]
+  --
   -- @since 1.2.0
   uniformM :: StatefulGen g m => g -> m a
 
+  default uniformM :: (StatefulGen g m, Generic a, GUniform (Rep a)) => g -> m a
+  uniformM = fmap to . (`runContT` pure) . guniformM
+
+class GUniform f where
```
Shimuuar, line 526: Use of ContT is nonobvious so it would be nice to add comment about it

```suggestion
-- | Default implementation of Uniform type class in terms
-- generics. Note use of ContT. Without it uses of fmap and bind are
-- fully polymorphic and GHC can't inline or specialize it. ContT
-- makes fmap and binds used in @gniformM@ monomorphic so GHC able to
-- specialize code instance reasonably close to handwritten one
class GUniform f where
```

src/System/Random/GFinite.hs
Bodigrim: Fixed.

Bodigrim:
@Bodigrim @lehins @idontgetoutmuch @curiousleo any more review suggestions? 
(I'm happy to remove `UniformRange` instance until we sort out `UniformRange` for tuples)

lehins:
> I'm happy to remove UniformRange instance until we sort out UniformRange for tuples)

@Bodigrim Yes, please. The instance that gets derived for tuples in this PR is not something we want in this library. I don't care if it follows all the laws in the World, but `Ord` instance on tuples should not get anywhere near `UniformRange` class.

Bodigrim:
@lehins done.

test/Spec.hs
```diff
@@ -141,3 +145,26 @@ runSpec = testGroup "runGenState_ and runPrimGenIO_"
 -- | Create a StdGen instance from an Int and pass it to the given function.
 seeded :: (StdGen -> a) -> Int -> a
 seeded f = f . mkStdGen
+
+#if __GLASGOW_HASKELL__ >= 802
```
lehins, line 149: It might be a dumb question. I'd be surprised if that generic deriving mechanism doesn't work for older ghc, so why this  CPP is needed here?

lehins (approved):
I'll admit that am far from a Generics guru, but overall it seems legit and of course the functionality it provides is incredibly useful. :+1: from me. @Shimuuar you had a look at this PR before, is there anything you'd like to add before we merge it?

test/Spec.hs
Shimuuar: AFAIR Generics were introduced in 7.2

Shimuuar:
No everything is good

test/Spec.hs
Bodigrim: GHC 8.2 improves UX of `{-# LANGUAGE DeriveAnyClass  #-}`, while older GHCs for some reason require `Uniform ()` instance and ask to put `instance Uniform MyAction` on a separate line. I think adding `Uniform ()` is a good idea anyways, so I did this and removed CPP in f009be5.

PR #72, opened 2020-06-29
Tuple instances random

lehins:
Fix for: #26 

Instances for tuples. Improved coverage. Fixes to `Random` haddock.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@
 * Addition of `runStateGenST_`
 * Ensure that default implementation of `ShortByteString` generation uses unpinned memory.
 * Addition of `TGen` and `TGenM`
+* Addition of tuple instances for `Random` up to 7-tuple
 
 # 1.2.0
 
--- a/src/System/Random.hs
+++ b/src/System/Random.hs
@@ -60,6 +60,7 @@ module System.Random
 
 import Control.Arrow
 import Control.Monad.IO.Class
+import Control.Monad.State.Strict
 import Data.ByteString (ByteString)
 import Data.Int
 import Data.IORef
@@ -316,12 +317,12 @@ instance Random CUIntPtr
 instance Random CIntMax
 instance Random CUIntMax
 instance Random CFloat where
-  randomR (CFloat l, CFloat h) = first CFloat . randomR (l, h)
+  randomR r = coerce . randomR (coerce r :: (Float, Float))
   {-# INLINE randomR #-}
   random = first CFloat . random
   {-# INLINE random #-}
 instance Random CDouble where
-  randomR (CDouble l, CDouble h) = first CDouble . randomR (l, h)
+  randomR r = coerce . randomR (coerce r :: (Double, Double))
   {-# INLINE randomR #-}
   random = first CDouble . random
   {-# INLINE random #-}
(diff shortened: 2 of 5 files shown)


src/System/Random.hs
```diff
@@ -197,9 +200,15 @@ class Random a where
   -- | Takes a range /(lo,hi)/ and a pseudo-random number generator
   -- /g/, and returns a pseudo-random value uniformly distributed over the
```
Bodigrim, line 201: We should not promise that it is __uniformly__ distributed.

src/System/Random.hs
```diff
@@ -197,9 +200,15 @@ class Random a where
...
-  -- that the values /lo/ and /hi/ are ever produced, but they may be,
-  -- depending on the implementation and the interval.
+  -- what happens if /lo>hi/, but usually the values will simply get swapped.
+  --
+  -- For continuous types there is no requirement that the values /lo/ and /hi/ are ever
+  -- produced, but they may be, depending on the implementation and the interval.
+  --
+  -- There is no requirement to follow the @Ord@ instance and the concept of range can be
+  -- defined on per type basis. For example product types will treat their values
+  -- independently. In case when a lawful range is desired `uniformR` should be used
```
Bodigrim, line 210: It is a good place to elaborate what we mean by "independently" here. I think it's worth to give an example: "randomR (a,b) (c,d) generates tuples with the first component between a and c and the second component between b and d" or something along these lines.

src/System/Random.hs
lehins: I do agree that current wording isn't perfect, but that's a tricky one because it is uniform for most of the types. Maybe we can say instead:

> and for types that have a well defined range concept it returns a pseudo-random value uniformly distributed over the closed interval /[lo,hi]/...

Otherwise saying "closed interval" is not terribly accurate either. What does it mean for `(a, b)` to be inclusive, right?

Types that behave differently from "uniformly" are now documented, so it is less of a problem if we get it worded properly.

Bodigrim: I would probably write "For a tuple `(lo, hi)` and a pseudo-random number generator `g`, return a pseudo-random value between `lo` and `hi` (inclusive) together with a new generator".

lehins: Sorry for sidetracking and not clarifying myself. I only said

> What does it mean for (a, b) to be inclusive, right?

as a way to argue a point and it was more of a rhetorical question.

To be more precise I meant `rangeR ((alo, blo), (ahi, bhi)) g :: ((a, b), g)`. Clusivity with instances as they are defined in this PR only makes sense if `a` and `b` types are kept independent of each other.

src/System/Random.hs
```diff
@@ -186,8 +187,10 @@ genByteString :: RandomGen g => Int -> g -> (ByteString, g)
 genByteString n g = runStateGenST g (uniformByteStringM n)
 {-# INLINE genByteString #-}
 
--- | The class of types for which uniformly distributed values can be
--- generated.
+-- | The class of types for which random values can be generated. For most of the types
+-- instances of `Random` will produce values that are uniformly distributed on the full
```
curiousleo, line 191:
```suggestion
-- | The class of types for which random values can be generated. Most
-- instances of `Random` will produce values that are uniformly distributed on the full
```

src/System/Random.hs
```diff
@@ -186,8 +187,10 @@ genByteString :: RandomGen g => Int -> g -> (ByteString, g)
...
+-- range, but for those types that such range cannot be determined some sensible default
```
curiousleo, line 192:
```suggestion
-- range, but for those types without a well-defined "full range" some sensible default
```

CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
+# 1.2.1
```
curiousleo, line 1: Super-nit: while 1.2.1 is overwhelmingly likely to be the next version, my suggestion would be to accumulate changes under "Unreleased changes" and change the heading to "1.2.1" in the release commit.

```suggestion
# Unreleased changes
```

curiousleo: Thanks btw for adding this to the changelog - something I forgot to do in my PR.

src/System/Random.hs
Shimuuar: I think problem here is not "uniform", proposed randomR for tuples generate uniform distribution over N-dimensional rectangle. It conflict with "over the closed interval /[lo,hi]/". 

This is really question what is `Random`? What does it promise? Should `randomR` generate values in the *interval*? That will rule out instances for tuples. Should it generate values in some set specified by tuple (a,b) which exact meaning depends on type in question?

Shimuuar:
> UniformRange, which already has tuple instances.

UniformRange doesn't have tuple instances. And couldn't possibly have

curiousleo:
> > UniformRange, which already has tuple instances.
> 
> UniformRange doesn't have tuple instances. And couldn't possibly have

You are right of course. `Uniform` has tuple instances, `UniformRange` does not have tuple instances.

src/System/Random.hs
lehins: > This is really question what is Random? What does it promise?

The promise is that the values a random. Everything else is open to an interpretation on per type basis. 

> Should it generate values in some set specified by tuple (a,b) which exact meaning depends on type in question?

I think it is a reasonable assumption. A subset of possible values that is defined by a tuple in some sensible way. With exact documentation what that sensible way is.

lehins:
> I think it would be better for them to switch to using UniformRange, which already has tuple instances.

@curiousleo This has already been discussed in #26 and it was also pointed out here that `UnifromRange` can't have an instance for tuples

> I don't think we should add features related to Random. 

We've discussed this on multiple occasions. `Random` is not going away and there is nothing wrong if people want to continue using it. There are people from different backgrounds using Haskell, math gurus that care about laws and others who could care less about math and they just want to get some random values. As long as it is documented, there is nothing wrong with having sensible and useful instances that don't produce uniform values on the whole range or don't obey the ordering defined for them, because it is not always possible to define some combination of `Uniform` or `UniformRange`, here are some latest examples: tuples #26, `Complex` numbers #74, `Rational` #75 and `Natural` #73 
CC @idontgetoutmuch

Bodigrim:
@lehins since now we have instances for tuples both in `Uniform` and `UniformRange`, do you wish to pursue this PR further?

lehins:
I'll extract some tests from this PR and will use it to compare performance of derived vs manual instances for tuples. I'll keep it open just for the time being so I don't forget to do it.

lehins:
> @lehins since now we have instances for tuples both in Uniform and UniformRange, do you wish to pursue this PR further?

@Bodigrim After thinking for a while, yes, I believe it is a good idea to have tuples instances for `Random` as well.

src/System/Random.hs
lehins: I've added a couple of examples. I am not sure I have enough brainpower to elaborate what "independently" means. If anyone feels like improving the haddock we can always accept a PR later.

src/System/Random.hs
lehins: I don't think these instances are controversial in any way, moreover most programmers will likely infer exactly what those instances do without any more explanation. Resolving this discussion, anyone seeking an improvement to haddock can add a PR later on.

Issue #44, opened 2017-10-17
Add Random instance for Natural


ndmitchell:
With GHC 8.2 there is Natural in the base library, so it would be good if Random provided a suitable instance.

cartazio:
Excellent idea. . I’ve got a 1.2/2.0 branch in preparation.

But one must then ask: what distribution! There is no finite interval to
select uniformly for naturals or integers.  We could have it just be the
same range as a 64bit word or 128 bit word would be ... but that seems
unsatisfactory.

ndmitchell:
If you don't do what you do for `Integer` it would be massively confusing. And for `Integer` you use `maxBound :: Int` - which seems at least consistent.

treeowl:
@ndmitchell using `maxBound :: Int` sounds extremely wrong. I would expect something with more of a Gaussian-style shape with a fairly large standard deviation for `Integer`, and probably half of that for `Natural`.

ndmitchell:
@treeowl perhaps, all I'm saying is they should be consistent. I would like the random Integer to never overflow all my memory though, so it needs a limit of some kind.

treeowl:
Ah, I suppose it does, for such practical reasons. But `maxBound :: Int` seems way too small. If the standard deviation is something like  `maxInt ^ n` for some small `n` (to pick something pretty arbitrary), then you could cut off at some sensible multiple of that without losing much. What I really don't know is whether there *is* a good principled choice of distribution.

cartazio:
My current plan is to have the default natural interval be word 64.

Defaults are well, defaults.  And for infinite structures there’s no good
answer for computations that finish in finite time. Hence markov chains and
approximations algorithms

The main bottle neck at the moment is figuring out how to make some needed
systematic changes while making sure the breakages are all tolerable and
happen just once.

cartazio:
More color :

The current random instance / type class only supports / makes sense for
uniform distributions on finite width intervals that have a finite number
of points contained in those intervals.   Anything else can’t Be an
instance.

lehins:
There is a new `instance UniformRange Natural`. I don't know if adding `instance Random Natural` is a good idea. The broken instance for `Integer` is already there, so we'll keep it for backwards compatibility, but adding new broken instances doesn't feel right. In any case `Natural` can now be generated with `uniformR` and `uniformRM`

PR #73, opened 2020-06-29
Add instance Random Natural

Bodigrim:


lehins:
Dammit, I was sure I fixed that CPP stuff. It probably happened in some PR that never got merged. Oh well, I am really glad you noticed it.

There was discussion about `Random` instance for `Natrual` in #44 I was really on the edge of not defining this instance and even closed that issue, but since you already implemented it, I am OK with it to be included. After all that is exactly what `Random` class is for. :+1: 

Let's give it a day or so for someone to object, and we can merge it if no-one speaks up.

curiousleo:
Same comment as here: https://github.com/haskell/random/pull/72#pullrequestreview-439704449 - I don't think we should be adding features to `Random` at this point. There already is a `UniformRange Natural` instance. If anybody wants to generate `Natural` values in the `Word` range, they can already do it.

idontgetoutmuch:
I don't think we should add this. If someone wants to generate random values for `Natural` then they should use `UniformRange`.

PR #79, opened 2020-08-24
Remove tasty-expected-failure from dependencies

felixonmars:
Builds fine and all tests pass here.

### DIFF
--- a/random.cabal
+++ b/random.cabal
@@ -159,7 +159,6 @@ test-suite spec
         smallcheck >=1.2 && <1.3,
         tasty >=1.0 && <1.4,
         tasty-smallcheck >=0.8 && <0.9,
-        tasty-expected-failure >=0.11 && <0.13,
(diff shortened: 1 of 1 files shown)


Bodigrim:
It seems that we actually do not use `tasty-expected-failure` at all: no module imports `Test.Tasty.ExpectedFailure`. Could it be possibly removed from dependencies?

felixonmars:
> It seems that we actually do not use `tasty-expected-failure` at all: no module imports `Test.Tasty.ExpectedFailure`. Could it be possibly removed from dependencies?

Indeed. Updated :)

Bodigrim (approved):

lehins:
I also made a revision on hackage and removed the bounds on `tasty-expected-failure`. Removing the dependency itself is not possible with a revision, but there is no need for a redundant dependency getting in a way with its bounds. 

Thank you for catching this.

PR #83, opened 2020-11-30
Fix typo of runStateGen

unam3:


Bodigrim:
Nice catch! Could you please fix other two occurencies of `runGenState` as well? 
https://github.com/haskell/random/blob/bdb9b43bdd7d3c0b77762d383eea1da341537751/random.cabal#L45
https://github.com/haskell/random/blob/bdb9b43bdd7d3c0b77762d383eea1da341537751/test/Spec.hs#L142

unam3:
> Nice catch! Could you please fix two other occurencies of `runGenState` as well?
> 
> https://github.com/haskell/random/blob/bdb9b43bdd7d3c0b77762d383eea1da341537751/random.cabal#L45
> 
> https://github.com/haskell/random/blob/bdb9b43bdd7d3c0b77762d383eea1da341537751/test/Spec.hs#L142

doen

Bodigrim (approved):
(Travis build is stalled, because we are [out of credits](https://travis-ci.com/organizations/haskell/plan))

lehins:
Good catch. Considering it's a haddock only change I don't think we need to wait for travis. Merging

This raises however a different issue. Opened #84

PR #85, opened 2020-12-01
initialise github ci

chessai:

PR #86, opened 2020-12-02
Replace Travis CI with GitHub Actions CI

### LINKED ISSUE #84: Setup Github Actions
Travis is done with open source. Stalled builds are a major obstacle (see #83), time to switch to something else, github actions seems like a very reasonable alternative.

lehins:
Fixes #84 

Supercedes #85 

This PR ensures:

* Ubuntu builds for all major versions of ghc with `stack`
* MacOS and Windows builds for latest stable versions of ghc aslo with `stacl`, just to make sure we are OS agnostic
* We reuse ppa for getting ghc fast when we can and use stack to install it when we can't.
* We still retain doctest we still have retain coverage on coveralls (currently I am only aware a solution with `stack` on how to do this)

@Bodigrim Please check it out. One thing that is required however, is that someone from haskell org grabs a token from coveralls page and adds it as a `COVERALLS_TOKEN` secret to this repo. Until then we won't have coverage being reported. I think travis had some sort of integration with coveralls, which didn't require this token being supplied, doesn't work for Github Actions though.


.github/workflows/stack.yaml
```diff
@@ -0,0 +1,110 @@
...
+        sudo apt-get update
+        sudo apt-get install ghc-${{ matrix.ghc }}
+    - name: Install Stack and GHC
+      run: |
+        set -ex
+        curl -sSL https://get.haskellstack.org/ | sh -s - -f
+        curl -sSL https://gist.githubusercontent.com/lehins/fd36a8cc8bf853173437b17f6b6426ad/raw/4702d0252731ad8b21317375e917124c590819ce/git-modtime.hs -o git-modtime.hs
+        [ -n "${{ matrix.ghc }}" ] && [ "${{ matrix.os }}" == "ubuntu-latest" ] && STACK_ARGS="$STACK_ARGS --system-ghc"
+        [ -n "${{ matrix.stack-yaml }}" ] && STACK_YAML=${{ matrix.stack-yaml }}
+        stack --resolver ${{ matrix.resolver }} runghc git-modtime.hs
```
Bodigrim, line 95: It would be nice to add a comment, explaining what and why is going on.

.github/workflows/stack.yaml
```diff
@@ -0,0 +1,110 @@
...
+      env:
+        COVERALLS_TOKEN: ${{ secrets.COVERALLS_TOKEN }}
+      run: |
+        set -ex
+        [ -n "${{ matrix.ghc }}" ] && [ "${{ matrix.os }}" == "ubuntu-latest" ] && STACK_ARGS="$STACK_ARGS --system-ghc"
+        [ -n "${{ matrix.stack-yaml }}" ] && STACK_YAML=${{ matrix.stack-yaml }}
+        if [ "${{ matrix.os }}.${{ matrix.resolver }}" == "ubuntu-latest.lts-14" ] && [ -n "${COVERALLS_TOKEN}" ]; then
+          stack test random:spec random:legacy-test --coverage --haddock --no-haddock-deps
+          stack hpc report --all
+          curl -L https://github.com/lehins/stack-hpc-coveralls/releases/download/v0.0.6.0/shc.tar.gz | tar xz shc
```
Bodigrim, line 106: BTW is there something specific which blocks your fork of `stack-hpc-coveralls` from being merged upstream?

.github/workflows/stack.yaml
```diff
@@ -0,0 +1,110 @@
...
+      if: matrix.ghc != '' && matrix.os == 'ubuntu-latest'
+      run: |
+        sudo add-apt-repository ppa:hvr/ghc -y
+        sudo apt-get update
+        sudo apt-get install ghc-${{ matrix.ghc }}
+    - name: Install Stack and GHC
+      run: |
+        set -ex
+        curl -sSL https://get.haskellstack.org/ | sh -s - -f
+        curl -sSL https://gist.githubusercontent.com/lehins/fd36a8cc8bf853173437b17f6b6426ad/raw/4702d0252731ad8b21317375e917124c590819ce/git-modtime.hs -o git-modtime.hs
```
Bodigrim, line 92: I'd prefer to have this script inlined (maybe rewritten to Bash?). It feels uneasy to source it from a (potentially mutable) gist.

Bodigrim (commented):
Is there a reason to avoid `actions/setup-haskell`? My understanding is that it could take care about low-level details, but I never digged into its sources.

Bodigrim:
> One thing that is required however, is that someone from haskell org grabs a token from coveralls page and adds it as a `COVERALLS_TOKEN` secret to this repo.

Done.

.github/workflows/stack.yaml
lehins: I had it originally written in bash that worked on linux, but making it OS agnostic is a nightmare, so I decided instead to write in Haskell instead.

Here is a more detailed write up on this topic. I'll include the link to this answer in the CI yaml: https://stackoverflow.com/questions/60906336/stack-haskell-build-cache-of-source-files-with-github-actions/61178945#61178945

Source is immutable, because url contains the hash of the version, in other words the gist will probably change, but in order to get the updated version, the url will have to change.

.github/workflows/stack.yaml
lehins: Lack of maintainership, I guess. Note that until yesterday my fork was exactly the same as upstream. All I did was made a static binary release 0.0.5.0 that could be downloaded and used in CI (upstream release 0.0.4.0 doesn't work with newer stack versions), which I used for over a year now. Yesterday I added Github Actions support and made another release 0.0.6.0. 

We'll see maybe @rubik is still interested in keeping https://github.com/rubik/stack-hpc-coveralls up to date and make releases. I submitted a PR with updates here: https://github.com/rubik/stack-hpc-coveralls/pull/26

.github/workflows/stack.yaml
lehins: Will do. 

> Update modification time of files to the latest commit time that a file was modified in. This ensures that `stack` does not rebuild unchanged files. More info in this SO answer: https://stackoverflow.com/questions/60906336/stack-haskell-build-cache-of-source-files-with-github-actions/61178945#61178945

Do you think it's sufficient?

lehins:
@Bodigrim Thanks for the review. Let me know if you don't have any more questions and I'll merge this PR. Thank you also for setting the coveralls token, I just verified that it works.

Shimuuar:
I don't have any experience with guithub actions. Anything that works is good IMO

- Would it be possible to get GHC version into name of build? For people who don't use stack lts number means nothing, and doubt even stack users remember correspondence for all versions.

- Do we need to run CI on both push and PR? Effectively twice for PRs coming from this repo

- Just in case. Is there way to bust cache in case something breaks?

lehins:
> Would it be possible to get GHC version into name of build? For people who don't use stack lts number means nothing, and doubt even stack users remember correspondence for all versions.

This would be nice, however Github Actions creates names for build jobs using the variables specified in the build matrix, while stack cares about the resolver instead of ghc and can't use such a variable if we where to use it instead of a resolver. I am sure there could be some sort of hackery implemented to imitate the desired behavior, but I don't quite see a solution right now. If I ever figure it out I'll update the CI spec.

> Do we need to run CI on both push and PR? Effectively twice for PRs coming from this repo

Yes, we do need builds for PRs because otherwise we will not know if CI passes or fails for forked repos before we can merge it. And if we don't have builds for pushes then without a PR you don't know if the CI passes of fails for your regular development workflow.

Note, that most commonly work is done on a fork, which means that push build will run on a fork repo, while PR build will run on the source repo.

> Just in case. Is there way to bust cache in case something breaks?

Yes, it is a manual process of bumping up a version in github actions file: `jobs.build.env.cache-version`

Shimuuar:
Thanks! 

Maybe github picks first variable for build name and moving `ghc` on top of `resolver` will make github to pick it.

lehins:
@Shimuuar In order for github to pick `ghc` it will have to be at the top level:
```yaml
      matrix:
        # resolver: [nightly, lts-16, lts-14, lts-12, lts-11, lts-9, lts-6]
        ghc: [...]
        os: [ubuntu-latest]
```

but the script expects ghc version only when the system version of ghc can be used. Anyways. As you said, as long as it works it is not important. However, I'll keep my eyes open for a solution to this minor improvement next time I have to setup Github actions on some project. (I suspect `vector` will be the next one ;) )

.github/workflows/stack.yaml
Bodigrim: Yeah, I used yours `shc.tar.gz` as well. Good stuff, it would be nice to grab it directly from Hackage.

.github/workflows/stack.yaml
Bodigrim: `defaults: { run: { shell: bash } }` should make `bash` the default shell even for Windows build. Are there compatibility challenges nevertheless?

.github/workflows/stack.yaml
lehins: > it would be nice to grab it directly from Hackage.

I don't know. Building it from source vs a quick download of a static executable does not sound appealing to me. Especially when it is only needed in only one specific place. 

One way or another it does the exact job I need it for, so I am not too keen into putting much work into a tool like that.

.github/workflows/stack.yaml
lehins: Yes, even for MacOS

.github/workflows/stack.yaml
```diff
@@ -0,0 +1,114 @@
...
+    branches: [ master ]
+
+defaults: { run: { shell: bash } }
+
+jobs:
+  build:
+    name: CI
+    runs-on: ${{ matrix.os }}
+    strategy:
+      fail-fast: false
```
Bodigrim, line 16: Are you sure we do not want `fail-fast: true`?

Bodigrim:
We need at least one `cabal`-based build to build against latest dependencies and GHC. Stackage tends to lag behind quite a lot, e. g., even the latest Nightly still features `random-1.1`. Something simplistic will be enough.

```yaml
  build-cabal:
    runs-on: ubuntu-latest
    steps:
    - uses: actions/checkout@v2
    - uses: actions/setup-haskell@v1
      id: setup-haskell-cabal
      with:
        ghc-version: 'latest'
    - name: Update cabal package database
      run: cabal update
    - uses: actions/cache@v2
      name: Cache cabal stuff
      with:
        path: |
          ${{ steps.setup-haskell-cabal.outputs.cabal-store }}
          dist-newstyle
        key: ${{ runner.os }}-${{ matrix.ghc }}
    - name: Build and test
      run: |
        cabal sdist -z -o .
        cabal get random-*.tar.gz
        cd random-*/
        cabal test all
```

.github/workflows/stack.yaml
lehins: Yes, Ia m sure. I wanna see all failures, instead of the first one. It is always easy to cancel the whole build if quick termination is needed

Bodigrim: I'm concerned that in future GitHub Actions might become less generous for OSS, and we'll have to minimise build time. All right, let's worry about it later.

lehins:
@Bodigrim Yeah, that's a good idea. Thanks for supplying the snippet. Added.

Bodigrim (approved):
Well done!

.github/workflows/stack.yaml
lehins: > I'm concerned that in future GitHub Actions might become less generous for OSS

We'll go back to Travis-CI then :laughing: 

Let's not worry about it, Microsoft is rich, so let's just hope they won't get stingy.

lehins:
@Bodigrim Thank you for your feedback and suggestions!

PR #89, opened 2020-12-29
Untie random-1.1 from time

### LINKED ISSUE #88: Untie random-1.1 from time
I am working on refactoring `bytestring` test suite. It is currently organized as two packages. There is a `bytestring` package, which reside in a top-level folder, and a `bytestring-tests` package in `tests` sub-folder with the following configuration:

```cabal
test-suite {foo,bar,baz}
  hs-source-dirs:   . ..
  main-is:          {Foo,Bar,Baz}.hs
  other-modules:    Data.ByteString
                    Data.ByteString.Char8
                    Data.ByteString.Internal
                    Data.ByteString.Lazy
                    Data.ByteString.Lazy.Char8
                    Data.ByteString.Lazy.Internal
                    Data.ByteString.Short
                    Data.ByteString.Short.Internal
                    ...many more...
  build-depends:    base, ghc-prim, deepseq, random,
                    tasty, tasty-quickcheck, QuickCheck
```

So tests do not depend on `bytestring` package explicitly, but rather just happen to share source files. This is unfortunate for many reasons. For example, each test component recompiles all modules of `bytestring` anew. Tooling such as HLS cannot figure out which source file belongs to what. But most importantly this way we do not actually test `bytestring` package, because there are no guarantees that the configuration in `bytestring.cabal` (GHC flags, C sources, etc.) is the same.

An obvious solution is to abandon the hack with `hs-source-dirs: ..` and add `bytestring` as a proper `build-depends` of `bytestring-tests` package. This still does not work in a satisfactory manner, because now any change to `bytestring` causes rebuilding of `unix`, `random`, `directory`, `QuickCheck`, `process`, `optparse-applicative`, `tasty`, `tasty-quickcheck` (all of them depend on `bytestring` transitively) and, as you can imagine, it takes obnoxiously long for any sort of interactive development.

Ideally I would like to eliminate `bytestring` as a transitive dependency of the test suite. Not only it would enable us to run tests reasonably quick, but also the very reason for having a separate package `bytestring-tests` would disappear.

I patched `optparse-applicative` and `tasty` ecosystem so that they no longer depend on `bytestring`, when compiled with right flags. While `random-1.2` immediately depends on `bytestring`, tests (and `QuickCheck`) are happy to stick to `random-1.1`.

This endeavour was quite successful for Linux and MacOS builds, but there is a nasty complication on Windows. `random-1.1` depends on `time` as a source of entropy for the global random number generator. And on Windows platform `time` happen to depend on `Win32`, which depends on `bytestring`, which makes it a circular dependency.

While strategically I would like to deal somehow with `bytestring` dependency of `random-1.2+`, its next release does not look happening soon enough. I would like to untie `random-1.1` from `time` and release it as `random-1.1.0.1`. There are two possible courses of actions:

* One can make `time` an optional dependency of `random-1.1`, guarded by a Cabal flag. Users compiling with `time` disabled would lose some (but not all) bits of entropy in `theStdGen`. A patch takes only couple of lines: https://github.com/Bodigrim/random/commit/eb9f9c51c69bde50646e9943a591548f724ff2ea

* Or we can switch the source of entropy from `time` to `splitmix-0.1.0.3` (note that the latter does not normally depend on `time`). This way we there is no need for a new flag, and entropy becomes better than in `random-1.1`. Patch is also very small: https://github.com/Bodigrim/random/commit/1686804d8f40d77c8bcdf5ebefda2edf618a8e41

I'm mildly in favor of the second approach, because it improves and does not complicate things.

Bodigrim:
Closes #88.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,6 @@
+# 1.1.0.1
+make `time` an optional dependency
+
 # 1.1
   * breaking change to `randomIValInteger` to improve RNG quality and performance
     see https://github.com/haskell/random/pull/4 and
--- a/System/Random.hs
+++ b/System/Random.hs
@@ -1,6 +1,10 @@
 #if __GLASGOW_HASKELL__ >= 701
 {-# LANGUAGE Trustworthy #-}
 #endif
+#if !defined(__NHC__) && !defined(TIME)
+{-# LANGUAGE ForeignFunctionInterface #-}
+#endif
+
 
 -----------------------------------------------------------------------------
 -- |
@@ -88,7 +92,11 @@ import Foreign.Ptr      ( Ptr, nullPtr )
 import Foreign.C	( CTime, CUInt )
 #else
 import System.CPUTime	( getCPUTime )
+#ifdef TIME
 import Data.Time	( getCurrentTime, UTCTime(..) )
+#else
+import Foreign.Ptr      ( Ptr, nullPtr )
+#endif
 import Data.Ratio       ( numerator, denominator )
 #endif
 import Data.Char	( isSpace, chr, ord )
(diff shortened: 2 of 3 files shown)


Bodigrim:
Ah, it does not solve my conundrum with `bytestring`, because there appear to be other packages (like, `ansi-terminal`) depending on `Win32`. I should really get a Windows box to experiment locally. 

Since otherwise there is little motivation to bother, closing.

PR #90, opened 2020-12-30
Allow tasty-1.4

Bodigrim:


lehins:
Thanks, made a revision on Hackage too. Also removed `Bodigrim-patch-1` branch

PR #91, opened 2021-01-18
Add special stack.yaml for nightly and update CI

lehins:


stack.yaml
```diff
@@ -1,7 +1,6 @@
-resolver: lts-15.1
+resolver: lts-16.27
 packages:
 - .
 extra-deps:
-- splitmix-0.1@sha256:d50c4d0801a35be7875a040470c09863342514930c82a7d25780a6c2efc4fda9,5249
-- doctest-0.17@sha256:684b9736321c255fc8da651a817a14d46afd9aa08a753b344abaa77b343123f2,6988
+- splitmix-0.1.0.3@sha256:fc3aae74c467f4b608050bef53aec17904a618731df9407e655d8f3bf8c32d5c,6049
 - smallcheck-1.2.0@sha256:8b431572e6a0503223e0e52014d41084c1b01f2aeea3bd499f6f529b3f6dfa89,1482
```
Bodigrim, line 6: Could you please update it either to `smallcheck-1.2.1` or to the latest revision of `smallcheck-1.2.0`?

lehins: Sure, but do you mind me asking why does it matter? It works with older version just fine, doesn't it?

Bodigrim: Just in case if someone changes a resolver to `nightly`, without pointing `stack` to `stack-nightly.yaml`.

lehins: Fair enough :smiley: It's done.

PR #94, opened 2021-01-27
WIP: Implement safetey around non-splittable PRNGs

lehins:
I think I figured out an elegant solution for differentiating splittable vs non-splittable PRNGs in a type safe way with very little breakage!!!

CC @idontgetoutmuch and @curiousleo We've discussed the problem here at some lengths: idontgetoutmuch/random#7

This is the change to the type class:

```haskell
class RandomGen g where
  type Splittable g :: Bool
  type Splittable g = 'True
  ...
  split :: Splittable g ~ 'True => g -> (g, g)
```

Which means only instances for PRNGs that are non splittable need to be adjusted, eg:
```haskell
instance RandomGen PureMT where
   type Splittable PureMT = TypeError ('Text "PureMT is not a splittable PRNG")
   split = error "System.Random.Mersenne.Pure: unable to split the mersenne twister"
```

which if you try to use `split` function, will give you a nice error message:

```haskell
impossibleSplit :: IO (PureMT, PureMT)
impossibleSplit = split <$> newPureMT
```

```
    • PureMT is not a splittable PRNG
    • In the first argument of ‘(<$>)’, namely ‘split’
      In the expression: split <$> newPureMT
      In an equation for ‘impossibleSplit’:
          impossibleSplit = split <$> newPureMT
   |
Compilation failed.
```

Instead of this ugly runtime error:
```
*** Exception: System.Random.Mersenne.Pure: unable to split the mersenne twister
CallStack (from HasCallStack):
```

### DIFF
--- a/src/System/Random/Internal.hs
+++ b/src/System/Random/Internal.hs
@@ -1,5 +1,6 @@
 {-# LANGUAGE BangPatterns #-}
 {-# LANGUAGE CPP #-}
+{-# LANGUAGE DataKinds #-}
 {-# LANGUAGE DefaultSignatures #-}
 {-# LANGUAGE FlexibleContexts #-}
 {-# LANGUAGE FlexibleInstances #-}
@@ -84,6 +85,7 @@ import Foreign.Storable (Storable)
 import GHC.Exts
 import GHC.Generics
 import GHC.IO (IO(..))
+import GHC.TypeLits
 import GHC.Word
 import Numeric.Natural (Natural)
 import System.IO.Unsafe (unsafePerformIO)
@@ -110,6 +112,9 @@ import Data.ByteString (ByteString)
 {-# DEPRECATED next "No longer used" #-}
 {-# DEPRECATED genRange "No longer used" #-}
 class RandomGen g where
+  type Splittable g :: Constraint
+  type Splittable g =
+    TypeError ('ShowType g ':<>: 'Text " is not a splittable RandomGen")
   {-# MINIMAL split,(genWord32|genWord64|(next,genRange)) #-}
   -- | Returns an 'Int' that is uniformly distributed over the range returned by
   -- 'genRange' (including both end points), and a new generator. Using 'next'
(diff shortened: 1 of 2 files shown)


curiousleo:
That looks _very_ nice!

It does seem to [break `deriving`](https://github.com/haskell/random/runs/1779622792#step:8:889) on old LTS. But that looks fixable. Other than that, I don't see any downsides with this approach.

It may be worth circulating this on the libraries list: (1) this technique deserves to be more well-known, (2) perhaps the authors of PRNG implementations have feedback.

Shimuuar:
That's cute hack!

I don't think it's possible without dropping either type family or GHC<8.2. Prior to 8.2 GHC could derive newtype instances for classes with associated types. See https://gitlab.haskell.org/ghc/ghc/-/issues/2721

lehins:
We can manually provide those instances, they don't have to be derived.

idontgetoutmuch:
Nice :)

lehins:
I figured how to improve it even further.
```haskell
class RandomGen g where
  type Splittable g :: Constraint
  type Splittable g =
    TypeError ('ShowType g ':<>: 'Text " is not a splittable RandomGen")
  ...
  split :: Splittable g => g -> (g, g)
```
Any generator instance that is actually splittable only has to override the `TypeError` with `()`:

```haskell
instance RandomGen SM.SMGen where
  type Splittable SM.SMGen = ()
  ...
  split = SM.splitSMGen
```

Here what we get in ghci:

```haskell
> data NoopGen = NoopGen
> :{
| instance RandomGen NoopGen where
|   genWord64 _ = (0, NoopGen)
|   split _ = error "Impossible"
| :}
> split NoopGen

<interactive>:8:1: error:
    • NoopGen is not a splittable RandomGen
    • In the expression: split NoopGen
      In an equation for ‘it’: it = split NoopGen
```

CC @treeowl You've expressed interest in this in #97, what do you think about this approach?


Naturally, this is a backwards incompatible solution, but was always meant to be a breaking change. It can be included in `random-1.3.0` whenever that might happen.

lehins:
Closing in favor of #160

PR #95, opened 2021-01-28
Implementation of TGenM for STM

lehins:
I don't know why I haven't thought of this before, but here is a mutable generator that works in STM: `TGenM` that is based on a `TVar`

### DIFF
--- a/src/System/Random/Stateful.hs
+++ b/src/System/Random/Stateful.hs
@@ -67,6 +67,12 @@ module System.Random.Stateful
   , applySTGen
   , runSTGen
   , runSTGen_
+  -- ** Mutable adapter in 'STM'
+  , TGen(..)
+  , TGenM(..)
+  , newTGenM
+  , newTGenMIO
+  , applyTGen
 
   -- * Pseudo-random values of various types
   -- $uniform
@@ -101,6 +107,7 @@ module System.Random.Stateful
 import Control.DeepSeq
 import Control.Monad.IO.Class
 import Control.Monad.ST
+import GHC.Conc.Sync (STM, TVar, newTVar, newTVarIO, readTVar, writeTVar)
 import Control.Monad.State.Strict
 import Data.IORef
 import Data.STRef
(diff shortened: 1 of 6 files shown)


lehins:
Anyone feels like reviewing this PR? If not, I'll merge it over the weekend and make it part of the upcoming release #110

src/System/Random/Stateful.hs
```diff
@@ -552,6 +567,73 @@ runSTGen_ :: RandomGen g => g -> (forall s . STGenM g s -> ST s a) -> a
...
+
+-- | Applies a pure operation to the wrapped pseudo-random number generator.
+--
+-- ====__Examples__
+--
+-- >>> import Control.Concurrent.STM
+-- >>> import System.Random.Stateful
+-- >>> let pureGen = mkStdGen 137
+-- >>> stmGen <- newTGenMIO pureGen
+-- >>> atomically $ applyTGen uniform stmGen :: IO Int
```
Shimuuar, line 626: It's better to use fixed size number here. I've been bitten by it: https://github.com/haskell/mwc-random/issues/79

Shimuuar (approved):
Looks good. It adds dependency on STM but since it's de facto part of stanadrd library it's not a problem. 

I thought it could cause problems for STM if its test suite depends on random and thus becomes nonbuildable. But it turns out stm does not have one, at least in cabal file

src/System/Random/Stateful.hs
lehins: Good point! Thank you. Switched to `Int32`

lehins:
>  It adds dependency on STM but since it's de facto part of stanadrd library it's not a problem.

Actually only the test suite depends on `stm`. The implementation itself imports `TVar` directly from base: `import GHC.Conc.Sync`:
https://github.com/haskell/random/pull/95/files#diff-2c181ac3f73b5e3eadd5b0019d00d2a54ecbcc244a8d583a8d2aeb1cad97eda1R110

PR #71, opened 2020-06-28
Allow deriving Uniform and UniformRange instances using DerivingVia

lambdadog:
This solution was proposed by @Shimuuar on [#21](https://github.com/haskell/random/issues/21#issuecomment-649094523).

Essentially, this should allow users of the random library to get `Uniform` and `UniformRange` instances for free, using `DerivingVia`

## Usage (basically copied from @Shimuuar's comment):

```haskell
{-# LANGUAGE DerivingVia #-}

data RGB = R | G | B
  deriving (Enum, Bounded)
  deriving Uniform via UniformEnum RGB
```

PR #96, opened 2021-01-29
make Uniform/UniformRange newtype-derivable

amesgen:
Allows to newtype-derive `Uniform`/`UniformRange` on GHC 8.6+, using `QuantifiedConstraints` (similar to [this](https://ryanglscott.github.io/2018/03/04/how-quantifiedconstraints-can-let-us-put-join-back-in-monad/)).

Also allows `DerivingVia`, applicable to #92. As an example, we revive #71.

 - For some reason, it only works with `StandaloneDeriving`.
 - `RepLike` might not be the best name. Should `RepLike` be exported, even though that is not strictly necessary?
 - On GHC < 8.6, there will be "unused import" warnings about `StateT` (we need the constructor to be available for `Coercible` instances). Could resolve with more CPP.
 - Missing documentation (will add if desired).

Feel free to close this if you think this is not worth it.

### DIFF
--- a/random.cabal
+++ b/random.cabal
@@ -155,6 +155,7 @@ test-suite spec
     build-depends:
         base -any,
         bytestring -any,
+        mtl,
         random -any,
         smallcheck >=1.2 && <1.3,
         tasty >=1.0 && <1.5,
--- a/src/System/Random.hs
+++ b/src/System/Random.hs
@@ -27,6 +27,7 @@ module System.Random
   , Random(..)
   , Uniform
   , UniformRange
+  , UniformEnum(..)
   , Finite
 
   -- ** Standard pseudo-random number generator
@@ -59,6 +60,7 @@ module System.Random
 
 import Control.Arrow
 import Control.Monad.IO.Class
+import Control.Monad.State.Strict (StateT(..))
 import Data.ByteString (ByteString)
 import Data.Int
 import Data.IORef
--- a/src/System/Random/Internal.hs
+++ b/src/System/Random/Internal.hs
@@ -20,6 +20,10 @@
 {-# LANGUAGE TypeFamilies #-}
 {-# LANGUAGE KindSignatures #-}
 #endif
+#if __GLASGOW_HASKELL__ >= 806
+{-# LANGUAGE ConstraintKinds #-}
+{-# LANGUAGE QuantifiedConstraints #-}
+#endif
 {-# OPTIONS_HADDOCK hide, not-home #-}
 
 -- |
@@ -33,6 +37,7 @@
 module System.Random.Internal
   (-- * Pure and monadic pseudo-random number generator interfaces
     RandomGen(..)
+  , RepLike
   , StatefulGen(..)
   , FrozenGen(..)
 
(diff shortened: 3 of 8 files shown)


Bodigrim:
I do not particularly like polluting all classes with `RepLike m` constraint. It's not particularly restricting, just unsure if it is worth it. 

Defining trivial `RepLike` for GHC < 8.6, which does not help to derive `instance UniformRange`, could be very much confusing for users. It could be easier to drop support of old GHCs entirely than provide a conditional API.

Bodigrim:
There is already a default definition of `instance Uniform` which works fine for enumeratable data:
https://github.com/haskell/random/blob/46a15afe68db473523364bb0573884d6f6b5406e/src/System/Random/Internal.hs#L523-L524
And #92 brings the same functionality for `UniformRange`. 

On the other hand, using `Enum` interface when possible is faster (both in compile and in run time) than `Generic`.

amesgen:
> I do not particularly like polluting all classes with RepLike m constraint. It's not particularly restricting, just unsure if it is worth it.

Yeah, it is certainly not elegant. In a better™ world, `forall a b. Coercible a b => Coercible (f a) (f b)` would be a superclass constraint on `Functor` (see e.g. [here](https://oleg.fi/gists/posts/2019-07-31-fmap-coerce-coerce.html#conclusion) for an argument).

> Defining trivial `RepLike` for GHC < 8.6, which does not help to derive `instance UniformRange`, could be very much confusing for users. It could be easier to drop support of old GHCs entirely than provide a conditional API.

Note that if one is only interested in < GHC 8.6 (e.g. one has an application on GHC 8.4) one can simply ignore that `RepLike` exists. You only have to worry about `RepLike` if you also want to support GHC 8.6+.

---

One thing that would need more exploration is whether `DerivingVia` would allow customization of the derived instances (e.g. different probabilities for different constructors), similar to [`deriving-aeson`](https://github.com/fumieval/deriving-aeson#deriving-aeson).

---

That said, I am also still not sure if this is worth it. The `Generic` instances are very nice as an alternative, but you are right that they are less performant (see e.g. https://gitlab.haskell.org/ghc/ghc/-/issues/5642 for compile time issues, deriving `(Ord, Eq, Enum, Bounded)` for the`BigSum` type is instant for me).

Bodigrim:
> Note that if one is only interested in < GHC 8.6 (e.g. one has an application on GHC 8.4) one can simply ignore that `RepLike` exists. 

If you are an app developer with GHC 8.4 then you see `UniformEnum` in the documentation, the symbol itself is available - but it does not work, you receive a cryptic error about roles. 

----

How about a low-tech solution? We can just export

```haskell
uniformEnumM :: forall g m a. (Enum a, Bounded a) => StatefulGen g m => g -> m a 
uniformEnumM g = pure . toEnum =<< uniformRM (fromEnum (minBound :: a), fromEnum (maxBound :: a)) g
```

so that clients could write 

```haskell
instance Uniform Things where uniformM = uniformEnumM
``` 

This is two-symbols shorter than 

```haskell
deriving via UniformEnum Things instance Uniform Things
```

No fancy extensions involved, no new type classes sticking out of everything.

amesgen:
Exporting `uniformEnumM` (and probably also `uniformEnumRM`) is a very good idea.

Also, I now lean in favor of closing this PR, and revisiting it when `random` has dropped support for GHC < 8.6, as the possible confusion right now is probably not worth the cost.

lehins:
> instance Uniform Things where uniformM = uniformEnumM

I also very much prefer this approach.

> I now lean in favor of closing this PR

Yes, feel free to close it. I don't think all the complexity is really worth it

PR #99, opened 2021-04-02
Support GHC 9.2

kakkun61:
Tested with GHC 9.2.1 α1.


andrewthad:
Not that there's anything that could have been done differently in this PR, but it would be nice if GHC provided a `ordW32#` so that

    charToWord32 (C# c#) = W32# (wordToWord32# (int2Word# (ord# c#)))

could become

    charToWord32 (C# c#) = W32# (ordW32# c#)

so that during codegen, GHC wouldn't have to try to get rid of the "mask out the upper bits" operation that `wordToWord32#` normally needs to perform.

PR #101, opened 2021-04-16
Fix performance ghc 9 and bench

lehins:
Benchmarks now consume the resulting generator produced by computation that is being benchmarks, thus it can no longer be optimized away by the ghc. Fixing benchmarks brings them to more realistic and consistent runtimes.

Fixed benchmarks still pointed out to regressions due to worse inliner heuristics in ghc-9 In order to workaround that problem we just need to add INLINE for all functions and just call it a day. Which is also done  in this PR

Getting rid of problems with inliner we are still left with a few regressions with ghc-9, albeit not as drastic (only ~60-70%). I think it will be good to follow up on those with a separate investigation another PR.

### DIFF
--- a/bench/Main.hs
+++ b/bench/Main.hs
@@ -16,6 +16,9 @@ import Test.Tasty.Bench
 
 import System.Random.Stateful
 
+seed :: Int
+seed = 1337
+
 main :: IO ()
 main = do
   let !sz = 100000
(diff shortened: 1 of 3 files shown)


Bodigrim:
Sorry, I'm AFK until Tuesday, so cannot take a close look. 

Let's separate making benchmarks meaningful and forcing inlining into separate PR. The reason is that I've had a similarly looking inlining regression in `text` and there is an intention to fix this on GHC side. We can opt for forcing inlining anyways later, but for now it would be useful to provide GHC team with unamended source code. 

See https://gitlab.haskell.org/ghc/ghc/-/issues/19557 and https://gitlab.haskell.org/ghc/ghc/-/merge_requests/5547

----

Does the claim about 1000x speed up remains valid?

Is it possible to setup `env` once and thread the same generator through all benchmarks?

lehins:
> Sorry, I'm AFK until Tuesday, so cannot take a close look at the moment.

Take you time, I don't expect mass adoption of ghc-9 by Tuesday :smile: 

> Let's separate making benchmarks meaningful and forcing inlining into separate PRs.

They are already implemented in two separate commits, which is good enough for separating the concerns.

> Does the claim about 1000x speed up remains valid?

Yes, that claim did not use those benchmarks but instead we used legacy benchmarks that were included in old random as well. I personally didn't even use this benchmark suite for anything because I had my own repo that I used when working on random. 
(I prefer criterion over gauge, so I am really happy with this switch to tasty-bench ;)

> Is it possible to setup env once and thread the same generator through all benchmarks?

Possible, but it doesn't make a difference. And if it doesn't make a difference, I am not gonna waste my time on doing the work.

bench/Main.hs
```diff
@@ -24,49 +27,63 @@ main = do
...
+      , env (pure $ SM.mkSMGen $ fromIntegral seed) $ \smGen ->
+          bench "split" $ whnf (genMany SM.splitSMGen smGen) sz
       ]
     , bgroup "pure"
       [ bgroup "random"
-        [ pureRandomBench (Proxy :: Proxy Float) sz
-        , pureRandomBench (Proxy :: Proxy Double) sz
-        , pureRandomBench (Proxy :: Proxy Integer) sz
+        [ env (pure (mkStdGen seed)) $ \gen ->
+            bench "Word32'" $ whnf (genMany (random :: StdGen -> (Word32, StdGen)) gen) sz
```
Bodigrim, line 42: Why do we need this additional benchmark?

lehins: We don't. I just forgot to get rid of it.

Bodigrim (approved):

PR #100, opened 2021-04-14
Switch from gauge to tasty-bench

Bodigrim:
`gauge` still cannot be compiled with GHC 9.0 because of `basement`, and further it will be broken once again by sized primitives in GHC 9.2. Switching to `tasty-bench` allows running benchmarks against GHC 9.0 and 9.2, which reveals pretty gruesome picture. 

```sh
cabal bench -w ghc-8.10.4 --ghc-options '-fproc-alignment=64' --benchmark-options '--csv 8.10.4.csv --hide-successes' random:bench
```

1. GHC 8.10.4 vs. GHC 9.0.1:

```sh
cabal bench -w ghc-9.0.1  --ghc-options '-fproc-alignment=64' --benchmark-options '--baseline 8.10.4.csv --csv 9.0.1.csv --hide-successes --fail-if-slower 50' random:bench
```

```
All
  pure
    uniformR
      full
        Word:                       FAIL (0.26s)
          488 μs ±  22 μs, 879% slower than baseline
        Int:                        FAIL (0.13s)
          485 μs ±  43 μs, 885% slower than baseline
        Char:                       FAIL (0.18s)
           25 ms ± 1.7 ms, 8733% slower than baseline
      excludeMax
        Char:                       FAIL (0.39s)
           25 ms ± 2.2 ms, 6210% slower than baseline
      includeHalf
        Char:                       FAIL (0.18s)
           26 ms ± 1.6 ms, 7748% slower than baseline
      floating
        St
          uniformFloat01M:          FAIL (0.16s)
            623 μs ±  45 μs, 1169% slower than baseline
          uniformFloatPositive01M:  FAIL (0.33s)
            642 μs ±  22 μs, 1206% slower than baseline
          uniformDouble01M:         FAIL (0.17s)
            626 μs ±  44 μs, 1157% slower than baseline
          uniformDoublePositive01M: FAIL (0.17s)
            631 μs ±  46 μs, 1177% slower than baseline

9 out of 40 tests failed (9.74s)
```

2. GHC 8.10.4 vs GHC 9.2.0 alpha:

```sh
cabal bench -w ghc-9.2.0.20210331 --allow-newer='split:base,splitmix:base,tagged:template-haskell' --ghc-options '-fproc-alignment=64' --benchmark-options '--baseline 8.10.4.csv --csv 9.2.0.csv --hide-successes --fail-if-slower 50' random:bench
```

```
All
  pure
    uniformR
      full
        Word:                       FAIL (0.24s)
[...]

PR #102, opened 2021-04-28
Inspection testing?

Bodigrim:
How do we feel about inspecting Core in our tests by means of [`inspection-testing`](http://hackage.haskell.org/package/inspection-testing-0.4.5.0/docs/Test-Inspection.html)? This would for example allow to catch inlining regression in GHC 9.0 from #100.


lehins:
Personally I find benchmarks work best for catching performance regressions, which is exactly what caught them in #100 :wink: That being said I do still think that inspection testing is a great idea, but it would require someone actually writing those tests. So, if you are up to it, knock yourself out.

Shimuuar:
I think it's a great idea. Only problem is finding good invariants.

bgamari:
Thank you, @Bodigrim! As you know, I am very much a fan of `inspection-testing`. It is *much* easier to sort out why an `inspection-testing` test is failing than it is to sift through piles of Core and assembler after a benchmark regression.

lehins (approved):
LGTM. @Bodigrim feel free to merge when you think it is ready.

Bodigrim:
I do not quite understand why lts-14 build was fine in PR, but fails when merged. Might it be a caching issue?..

lehins:
@Bodigrim sorry didn't have time for OSS work during last month. Vacation and job change took all my focus away.

> I do not quite understand why lts-14 build was fine in PR, but fails when merged. 

As it turns out caching was not at fault. It is building with coverage that is causing the inspection tests to fail. LTS-14 is the only one that runs the coverage report and uploads to coveralls, that's why it didn't fail for PRs, but did when was merged. I've changed the CI script a bit to ensure that PRs go through the same build/test path. I'll open an issue about this, so we can discuss the best approach to solve/work around this inspection testing issue

PR #106, opened 2021-06-03
Reorder constraints for consistency and easy of type application 

### LINKED ISSUE #105: Better type variable ordering for type applications
I think it is more common to explicitly specify the type of thing that you want to generate than it is to specify the generator type, so I would propose to change the type variable ordering to make it easier to use explicit type applications, e.g. change the type signature of `uniform` from

```haskell
uniform :: (RandomGen g, Uniform a) => g -> (a, g)
```

To

```haskell
uniform :: (Uniform a, RandomGen g) => g -> (a, g) 
```

The particular use-case I have right now is using `fromIntegral` to convert the generated numbers to arbitrary length `Integer`s. I currently have to write:

```haskell
f :: StdGen -> (Integer, StdGen)
f g = first fromIntegral (uniform @_ @Word g)
```

But it would be nicer to be able to write:

```haskell
f :: StdGen -> (Integer, StdGen)
f g = first fromIntegral (uniform @Word g)
```

Are there any real-world examples where the current ordering would be preferred?

I also want to acknowledge that this is a breaking change and might be rejected for that reason alone, that would be sad but understandable.

noughtmare:
fixes #105

Should I change the `@since` annotations too?

### DIFF
--- a/src/System/Random.hs
+++ b/src/System/Random.hs
@@ -144,7 +144,7 @@ import qualified System.Random.SplitMix as SM
 -- (True,StdGen {unStdGen = SMGen 11285859549637045894 7641485672361121627})
 --
 -- @since 1.2.0
-uniform :: (RandomGen g, Uniform a) => g -> (a, g)
+uniform :: (Uniform a, RandomGen g) => g -> (a, g)
 uniform g = runStateGen g uniformM
 {-# INLINE uniform #-}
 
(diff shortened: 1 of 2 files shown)


lehins:
That is perfect!

> Should I change the `@since` annotations too?

No need. `@since` usually specifies the version  when the function was introduced and exported for public consumption. Any further changes to a function, even semantic changes, shouldn't affect it.

andrewthad:
This comment is intended more for the maintainers than for the OP. Consider an explicit forall to introduce the type variables. It makes it slightly more clear how TypeApplications will work with the function. I am not sure if there is already any president in this library for preferences around functions where use of TypeApplications is common.

lehins:
> Consider an explicit forall to introduce the type variables.

@andrewthad I would also prefer explicit `forall`, but unfortunately it is useless for class functions. And in this case we are talking about functions that are both inside a class (eg. `uniformM`) and standalone (eg., `uniform`). This makes it problematic for consistency. I wish ghc had a better story around `forall`, but IMHO, in this scenario avoiding explicit forall actually promotes consistency.

PR #109, opened 2021-06-06
Addition of `initStdGen`. Fix #103

### LINKED ISSUE #103: Expose method for initializing new StdGen from SplitMix.initSMGen
From #69 it seems that the preferred method for using the library is something like

```Haskell
main = do
  s <- round <$> getPOSIXTime
  g <- newIOGenM $ mkStdGen s
  -- whatever thing using randomness
  l <- replicateM 10 $ uniformRM (1 :: Int, 100) g
  print l
```

Using POSIXTime seems hacky, when there is a nicer method of getting a random seed:
```Haskell
import Control.Monad
import System.Random
import System.Random.Stateful
import System.Random.Internal
import qualified System.Random.SplitMix as SM

initStdGen :: IO StdGen
initStdGen = StdGen <$> SM.initSMGen

main = do
  g <- initStdGen >>= newIOGenM
  -- whatever thing using randomness
  l <- replicateM 10 $ uniformRM (1 :: Int, 100) g
  print l
```
But it isn't a good idea to import internal modules, so `initStdGen` has to be added to the API.

lehins:


curiousleo (approved):
LGTM!

Re the `random` changes: did you run a benchmark after the changes? Inlining works in strange ways. (Or is this perhaps already covered by inspection tests?)

lehins:
> Re the random changes: did you run a benchmark after the changes?

Yeah, no changes in performance.

PR #112, opened 2021-06-29
Avoid replicateM_ in benchmarks: GHC is too keen to throw it away

Bodigrim:
I was baffled by `uniformDouble01M` benchmarks:

```
$ cabal new-run random:bench:bench -- -t10 -p uniformR.floating.IO
All
  pure
    uniformR
      floating
        IO
          uniformFloat01M:          OK (5.28s)
            2.56 ms ± 220 μs
          uniformFloatPositive01M:  OK (1.28s)
            2.52 ms ± 225 μs
          uniformDouble01M:         OK (1.30s)
            39.2 μs ± 3.6 μs
          uniformDoublePositive01M: OK (2.54s)
            39.1 μs ± 1.7 μs
```

While having `uniformDouble01M` faster than `uniformFloat01M` is possible, it's hard to believe that it is 60x faster! It is also way too fast: we generate 100000 doubles, so the rate is 2.5 doubles per nanosecond. [On ARM machine](https://cloud.drone.io/Bodigrim/random/1/1/2) results are even more nonsensical:

```
All
  pure
    uniformR
      floating
        IO
          uniformFloat01M:          OK (0.27s)
            4.09 ms ± 256 μs
          uniformFloatPositive01M:  OK (0.11s)
            4.01 ms ± 374 μs
          uniformDouble01M:         OK (0.22s)
            770  ps ±  40 ps
          uniformDoublePositive01M: OK (0.20s)
            770  ps ±  40 ps
```

The reason is that our benchmarks use `replicateM_`, which basically gives GHC a green light to do nothing at all. Further outcome depends on inlining magic, and it seems that for `uniformFloat01M` the code at least iterates a generator (but does not produce any `Float`s), while for `uniformDouble01M` it's just an empty loop. Further, an empty loop is optimized by LLVM on ARM platform to a no-op instruction, giving us results in picoseconds range. 

The PR replaces `replicateM_` by a hand-written monadic loop, which takes care to force intermediate results. Now benchmarks make more sense:

```
All
  pure
    uniformR
      floating
        IO
          uniformFloat01M:          OK (1.67s)
            3.20 ms ± 228 μs
          uniformFloatPositive01M:  OK (3.28s)
            3.20 ms ± 201 μs
          uniformDouble01M:         OK (2.29s)
            2.25 ms ± 208 μs
          uniformDoublePositive01M: OK (2.27s)
            2.20 ms ± 148 μs
```

(Thanks to @csstaub, who helped me to identify the issue)

PR #113, opened 2021-06-30
Investigate benchmarks

lehins:
Benchmarks for `floating` point numbers seem suspicious, as it was rightly pointed out in #112 

This PR is designed to fix that issue by relying on writing generated random values into a `MutablePrimArray`, thus avoiding any potential ghc optimizations that can discard intermediate computations.

Before this PR:

```
      floating      
        IO          
          uniformFloat01M:          OK (0.80s)
            3.1 ms ± 194 μs
          uniformFloatPositive01M:  OK (0.19s)
            3.1 ms ± 260 μs
          uniformDouble01M:         OK (0.24s)
             58 μs ± 4.5 μs
          uniformDoublePositive01M: OK (0.18s)
             86 μs ± 6.3 μs
        pure        
          uniformFloat01M:          OK (0.16s)
            334 μs ±  25 μs
          uniformFloatPositive01M:  OK (0.19s)
            351 μs ±  26 μs
          uniformDouble01M:         OK (0.20s)
            760 μs ±  60 μs
          uniformDoublePositive01M: OK (0.21s)
            768 μs ±  54 μs
```

With this PR (only `floating/IO` has changed):

```
      floating
        IO
          uniformFloat01M:          OK (0.19s)
            370 μs ±  22 μs
          uniformFloatPositive01M:  OK (0.20s)
            382 μs ±  24 μs
          uniformDouble01M:         OK (0.20s)
            772 μs ±  55 μs
          uniformDoublePositive01M: OK (0.21s)
            801 μs ±  51 μs
        pure
          uniformFloat01M:          OK (0.17s)
            353 μs ±  24 μs
          uniformFloatPositive01M:  OK (0.76s)
            371 μs ±  19 μs
          uniformDouble01M:         OK (0.19s)
            750 μs ±  57 μs
          uniformDoublePositive01M: OK (0.19s)
            749 μs ±  72 μs
```

This PR also changes CI so older ghc versions are tested with cabal instead of stack, because new dependency `primitive-0.7.1.0` doesn't work with older Cabal the library, which is deeply ingrained into each LTS.


andrewthad:
Good work! I often use a similar approach in my own benchmarking, and it works well. The only thing to be careful of is that you really want to be sure that fillMutablePrimArrayM is inlined. The best way to do this is to annotate `fillMutablePrimArrayM` with `{-# inline fillMutablePrimArrayM #-}` and then to hoist all your tests into top-level non-inlined functions, turning this:

    , env ((,) <$> getStdGen <*> newAlignedPinnedPrimArray sz) $ \ ~(gen, ma) ->
        bench "uniformDouble01M" $
        nfIO (runStateGenT gen (fillMutablePrimArrayM uniformDouble01M ma))

Into this:

    , env ((,) <$> getStdGen <*> newAlignedPinnedPrimArray sz) $ \ ~(gen, ma) ->
        bench "uniformDouble01M" $
        nfIO (runStateGenT gen uniformDouble01M_specialized))
    ...
    uniformDouble01M_specialized :: StdGen -> ST s (PrimArray Double)
    {-# noinline uniformDouble01M_specialized #-}
    uniformDouble01M_specialized gen = fillMutablePrimArrayM uniformDouble01M ma gen

This makes the benchmarks more verbose, but it has several advantages for these kinds of low-level benchmarks:

1. The specialization is more reliable.
2. You can use `-ddump-simpl` and find `uniformDouble01M_specialized` and look for perf issues. If it's inlined into the massive blob of benchmark suite stuff, the dumped Core is difficult to find stuff in.
3. This prevents issues with code alignment, which can make a big difference in benchmarks that loops with short bodies. Read Andreas's response to my issue at https://gitlab.haskell.org/ghc/ghc/-/issues/16556 for a better explanation of this.

lehins:
Thank you @andrewthad for the review!

> The only thing to be careful of is that you really want to be sure that fillMutablePrimArrayM is inlined.

Yeah, I am quite familiar with this problem :) However, it is usually quite obvious from the benchmark results when a function is not inlined, its performance becomes terrible. Also, from what I know, when a function is defined in the same module, that it is being benchmarked in (i.e. it is where main is and it is not exported), then ghc will almost always inline it. So, I think `INLINE` pragma is is more of a concern when a function is imported from some other module.

>  hoist all your tests into top-level non-inlined functions

That is certainly an interesting suggestion, which I'll definitely have to take into consideration and experiment with. All the benefits you outlined make a lot of sense. However, I'll refrain from adding this approach in this PR, since, as far as I understand it, there should be little affect on the performance results. Most importantly though I have too much on my plate as it is at the moment to mess with this. I am pretty confident that the actual implementation of floating point number generation is pretty good, so I am not worried too much about those benchmarks.

andrewthad:
No worries. Just something to keep in mind if someone ever reports wobbly results across different versions of GHC.

Bodigrim (approved):
Nice!

PR #115, opened 2021-08-21
Added examples to randomIO and randomRIO

mpscholten:
This adds examples and improves documentation for `randomIO` and `randomRIO` :)

### DIFF
--- a/src/System/Random.hs
+++ b/src/System/Random.hs
@@ -390,13 +390,37 @@ getStdRandom f = liftIO $ atomicModifyIORef' theStdGen (swap . f)
   where swap (v, g) = (g, v)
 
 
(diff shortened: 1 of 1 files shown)


lehins:
@mpscholten Thank you for your PR. Documentation for  `randomIO` and `randomRIO` indeed needed some improvements.

Unfortunately, your suggested documentation additions do not align well with intentions of this library. First of all, both `randomIO` and `randomRIO` are historical baggage and are advised against, because global mutable state has huge downsides and unexpected consequences.  Second of all there is a new stateful interface that supersedes the approach of single global mutable PRNG. Lastly we also use doctests for all our examples.

I've tried to convey all of this in a new PR #119 ,  which I'd appreciate any feedback on. Closing this one in favor of the aforementioned one.

PR #120, opened 2021-09-13
Fix for inaccurate statement in the haddock about 32/64 bit architectures

lehins:
We recently where able to run the test suite on different architectures:

* 64bit s390x (Big Endian)
* 32bit Armv7,
*  and of course 32/64bit Intel x86_64/i386

A couple of new tests explicitly verify that the byte sequence produced by `StdGen` is architecture independent. These tests passed on all of those architectures. This gives me quite a bit of confidence that the statement "in order to get predictability the build has to be done on the same architecture (32-bit or 64-bit)" is not accurate. Please let me know if anyone thinks otherwise.

### DIFF
--- a/src/System/Random.hs
+++ b/src/System/Random.hs
@@ -517,11 +517,12 @@ randomIO = liftIO $ getStdRandom random
 -- $reproducibility
 --
 -- If you have two builds of a particular piece of code against this library,
--- any deterministic function call should give the same result in the two
--- builds if the builds are
+-- any deterministic function call should give the same result in the two builds
+-- if the builds are compiled against the same major version of this library.
 --
--- *   compiled against the same major version of this library
--- *   on the same architecture (32-bit or 64-bit)
+-- When using the same `StdGen` with deterministic functions, they will produce
+-- the same results independently of target architecture the program was build
+-- for: 32-bit or 64-bit, big endian or little endian.
 --
 -- $references
 --


Bodigrim:
@lehins I've been able to setup i386 and arm64 CIs on drone.io for another project. Do you think it's a good addition?

lehins:
> Do you think it's a good addition?

For a project like `random`, absolutely!

Bodigrim (approved):

src/System/Random.hs
```diff
@@ -517,11 +517,12 @@ randomIO = liftIO $ getStdRandom random
...
--- any deterministic function call should give the same result in the two
--- builds if the builds are
+-- any deterministic function call should give the same result in the two builds
+-- if the builds are compiled against the same major version of this library.
 --
--- *   compiled against the same major version of this library
--- *   on the same architecture (32-bit or 64-bit)
+-- When using the same `StdGen` with deterministic functions, they will produce
+-- the same results independently of target architecture the program was build
+-- for: 32-bit or 64-bit, big endian or little endian.
```
Shimuuar, line 525: I think this part is a bit misleading. This is true for splitmix (and any generator that generate randomness in 64-bit chunks). Generators that generate randomness in 32-bit chunks (mwc-random for example) will behave differently.

https://github.com/haskell/random/blob/66cdf501b0a80684ff882a6f7eff885f3ce7ce2b/src/System/Random/Internal.hs#L813-L819

Library do not provide guarantee that results will be same for 32/64 bit arch. StdGen just happens to work this way. And even then different width of Int/Word may bite you

lehins: This is exactly why documentation says: "When using the same `StdGen` ...".

I can add that something along the lines. The same is not necessarily true for other pseudo random number generators that implement `RandomGen`.

> And even then different width of Int/Word may bite you

I'd be great to see a concrete example of how it can bite.

I really don't want to tell users that they can have different results between architectures, if it is not true.

I'll add some more tests that verify this 32/64 bit compatibility claim.

lehins: I see what you  mean now. That `Uniform Int` instance will indeed produce different results on 32 vs 64 bits

Shimuuar: > I can add that something along the lines. The same is not necessarily true for other pseudo random number generators that implement RandomGen.

Yes I think it's better to tell about this explicitly. It's too easy to get impression that this applies generally when one isn't reading carefully which is usually the case.

> I'd be great to see a concrete example of how it can bite.

Even for splitmix `uniform` for `Word/Int` will produce different values. Wildly different since on one arch it only has 32 bits and on other 64. Then computation may have values that overflow 32 bit int but not 64. I don't have good example but 4e9 is not much. It's easy to go above this it you say counting bytes.

lehins: This  could be fixed with:

```haskell
 instance Uniform Int where 
   uniformM 
     | wordSizeInBits == 64 = 
       fmap (fromIntegral :: Word64 -> Int) . uniformWord64 
     | otherwise = 
       fmap ((fromIntegral :: Word32 -> Int) . (`shiftR` 32)) . uniformWord32 
```

but that would be a breaking change, so I'll open a ticket for this and close this PR for now.

lehins: Actually nevermind. Giving it a second though this compatibility is impossible and you are totally right. I am not gonna worry about this. Thank you Aleksey!

PR #128, opened 2022-03-08
Fix doctests using cabal-doctest

ysangkok:
Tested using

    cabal test doctests --enable-tests -w ghc-8.4.4

I have changed the mechanism used to disable these test to using `buildable: False`
instead of making the test just quit. That way, the user won't be misled
into thinking that the tests were actually run.

### DIFF
--- a/Setup.hs
+++ b/Setup.hs
@@ -1,4 +1,6 @@
-import Distribution.Simple
+module Main where
+
+import Distribution.Extra.Doctest (defaultMainWithDoctests)
 
 main :: IO ()
-main = defaultMain
+main = defaultMainWithDoctests "doctests"
--- a/random.cabal
+++ b/random.cabal
@@ -59,7 +59,7 @@ description:
     See "System.Random.Stateful" for more details.
 
 category:           System
-build-type:         Simple
+build-type:         Custom
 extra-source-files:
     README.md
     CHANGELOG.md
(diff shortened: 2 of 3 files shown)


ysangkok:
Oops, I did the rebase wrong, it is now fixed @lehins .

random.cabal
```diff
@@ -219,3 +219,8 @@ benchmark bench
         random,
         splitmix >=0.1 && <0.2,
         tasty-bench
+
+custom-setup
+ setup-depends:
+   base >= 4 && <5,
+   cabal-doctest >= 1 && <1.1
```
lehins, line 226: Why do you need cabal-doctest? It has been deprecated (see hackage)

Setup.hs
```diff
@@ -1,4 +1,6 @@
-import Distribution.Simple
+module Main where
+
+import Distribution.Extra.Doctest (defaultMainWithDoctests)
 
 main :: IO ()
-main = defaultMain
+main = defaultMainWithDoctests "doctests"
```
lehins, line 6: Custom set up with cabal-doctest has not been necessary for a few years now

lehins (changes requested):
I understand what your motivation is, but making doctests non buildable for ghc versions that don't support doctests simply complicates CI, without any real benefit for maintainers. In other words I am against this change, because doctests are primarily there for maintainers to ensure examples in the haddock haven't been broken. 

Thank you for your attempt though

lehins:
That is a very good question. The issue comes from the doctests needing ghc's env files. See how it is used on CI https://github.com/haskell/random/blob/c796015598f3eafaddaa9c79ec817ead3c55e936/.github/workflows/ci.yaml#L51-L52

In particular: `--write-ghc-environment-files=always`

So something like this should do the trick:

```shell
$ cabal configure --enable-tests
$ cabal build random:doctests --write-ghc-environment-files=always
$ cabal test random:doctests
```

PR #137, opened 2023-03-25
Update CI config, adding GHC 9.6 jobs

Bodigrim:
Inspection tests fail on GHC 9.6, because `M1` survives in types:

```haskell
        lvl_soCG
          :: M1
               C
               ('MetaCons "Never" 'PrefixI 'False)
               (S1
                  ('MetaSel
                     'Nothing 'NoSourceUnpackedness 'NoSourceStrictness 'DecidedLazy)
                  (Rec0 Void))
               Any
        lvl_soCG
          = case error (lvl_soSv `cast` <Co:4> :: ...) lvl_soSx of wild_00 {
            }

        lvl_soTe :: MyAction
        lvl_soTe = Never (lvl_soCG `cast` <Co:43> :: ...)
```

This is quite irrelevant, `data MyAction = Code (Maybe Bool) | Never Void | Eat (Bool, Bool) | Sleep ()` and we do not really care about optimization of `error :: Void`, but I'm unsure what's the best way to mute this test.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -24,8 +24,9 @@ jobs:
           - { os: ubuntu-latest,  ghc: "8.8.4"  }
           - { os: ubuntu-latest,  ghc: "8.10.7" }
           - { os: ubuntu-latest,  ghc: "9.0.2"  }
-          - { os: ubuntu-latest,  ghc: "9.2.4"  }
-          - { os: ubuntu-latest,  ghc: "9.4.1"  }
+          - { os: ubuntu-latest,  ghc: "9.2.7"  }
+          - { os: ubuntu-latest,  ghc: "9.4.4"  }
+          - { os: ubuntu-latest,  ghc: "9.6.1"  }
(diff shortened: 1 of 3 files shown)


Bodigrim:
I think this is good to go now.

.github/workflows/ci.yaml
```diff
@@ -11,7 +11,7 @@ defaults: { run: { shell: bash } }
 jobs:
   build-cabal:
     name: CI-cabal
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-20.04
```
lehins, line 14: What's the reason for downgrading ubuntu? Latest version, as far as I know, is 22.04

lehins (approved):
Besides the ubuntu version, this PR looks great. Thank you!

.github/workflows/ci.yaml
Bodigrim: Reverted back to `ubuntu-latest`.

PR #138, opened 2023-04-23
CI: migrate away from drone.io

Bodigrim:
drone.io dropped free tier some time ago, so migrating i386 and arm jobs to GitHub.


.github/workflows/ci.yaml
```diff
@@ -199,3 +199,48 @@ jobs:
...
+        curl --proto '=https' --tlsv1.2 -sSf https://get-ghcup.haskell.org | BOOTSTRAP_HASKELL_NONINTERACTIVE=1 BOOTSTRAP_HASKELL_INSTALL_NO_STACK=1 sh
+    - uses: actions/checkout@v1
+    - name: Test
+      run: |
+        source ~/.ghcup/env
+        cabal update
+        cabal test
+
+  # We use github.com/haskell self-hosted runners for ARM testing.
+  arm:
```
lehins, line 221: This is really cool. However, current bobs have been queued for half a day. So, it looks like there is either something wrong with arm builders right now or there is too much contention on the self hosted arm machine resources. If it is the latter, maybe we should run them only on the merge commits to master? I would hate every PR to wait ages before it could get merged

Bodigrim: There is something wrong with setup: I do not see ARM runners at https://github.com/haskell/random/settings/actions/runners, while they are present for haskell/text, haskell/bytestring, haskell/filepath, etc. I might have forgotten how to get it right... @hasufell do you remember anything about it?

hasufell: Yes. The repository has to be added to the default runner group. We decided to make it a whitelist, so that we're aware who's using them.

I just added random to the group: https://github.com/organizations/haskell/settings/actions/runner-groups/1

Bodigrim: @lehins ARM jobs are good now. 

LTS 14 job fails with unrelated
```
/home/runner/work/_temp/60e7a972-e69e-4f10-b062-23c4d1d1f643.sh: line 17: 10951 Segmentation fault      (core dumped) ./shc --repo-token="$COVERALLS_TOKEN" --partial-coverage --fetch-coverage combined custom
```

Issue #139, opened 2023-07-27
Utilities to get a random element from a list and shuffle a list


subterfugue:
What I expect the signatures of these utilities to be:
1. `random-fu` implements the random element utility as `randomElement :: [a] -> RVar a`, though it doesn't have to be that specifically. Maybe a more `random` implementation would be `randomElement :: RandomGen g => [a] -> g -> a`?

2. `shuffle :: RandomGen g => [a] -> g -> [a]`.

I know that there was an issue precisely about this about 3 years ago, but honestly, I think this should be heavily rethought. 

These two functionalities are repeatedly provided and reimplemented in many other libraries, with a varying interface, which makes it all the more confusing whether it should be manually implemented by the user, or whether they should add in these libraries (which depend on `random` anyway) just to use these 2 utilities that aren't in `random`.

Forgive me if I'm making assumptions but this seems like this doesn't add many maintenance burdens, and makes dealing with randomness much easier than it is right now. This doesn't seem like it's introducing any sort of feature creep either, it's well within the scope of what `random` does, or at least to me. 

There are only a few criticisms of this in the other issue, mainly
1. Too opinionated
2. Performance and laziness
3. Already implemented by other libraries

For the first point, I think that being opinionated in this scenario is the better option than not. So far many libraries have implemented this and we have an overview of how performant/convenient different approaches are. This is aside from the fact that no matter what implementation we choose, we'll be giving a better experience for people than they're having right now. With pretty much an imperceptible number of downsides. If the opinion we make for the user doesn't suit them, then at that point they can make it on their own.

For the second point, I think this is something that most Haskellers have come to expect. Yes, bottoms and laziness will affect implementations. Yes, using linked lists is not recommended as the most performant collection. Even only providing it for linked list will, again, improve people's situations.

For the third point, `random` has already reimplemented behavior from other libraries. It seems to me way more than simply adding two utilities.

I'm not really a library maintainer, but as a Haskell programmer... To me, this addition is only an improvement, and it has no downsides.

bradrn:
I support this proposal. From discussion on the FP Discord server, it sounds like a number of other people support it too.

konsumlamm:
These two are the most basic random utilities I can think of and I'd expect any decent random library to have them.

I think it makes more sense to make `g` the first argument, for currying. As for the names, `shuffle` is a no-brainer imo, but perhaps `choose` instead of `randomElement` (the latter is rather verbose)?

lehins:
[...]

PR #140, opened 2023-07-27
Introduce `shuffleList` and `shuffleListM`

lehins:
As it was described in #139 this PR implements the most basic version of random list shuffling, both stateful and pure implementations. 

The way the shuffling is achieved is somewhat similar to the implementation in QuickCheck. There are better ways, but the one that I know of requires a mutable array, which would make a dependency footprint or implementation quite a bit more complicated.

We can experiment with performance and quality of alternative implementations for random-1.3, whenever that will come. Unless someone is willing to provide a better implementation (with some becnhmarks) soon enough as a PR into this branch, I think current implementation is sufficiently good enough for the initial version.

Changes in this PR are non-breaking, so I'll happily backport them into a minor `random-1.2.2` release.

As a nice extra this PR also contains a pure implementation of `uniformList`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,6 @@
 # 1.3.0
 
+* Add `shuffleList` and `shuffleListM`: [#140](https://github.com/haskell/random/pull/140)
 * Add `mkStdGen64`: [#155](https://github.com/haskell/random/pull/155)
 * Add `uniformListRM`, `uniformList`, `uniformListR`, `uniforms` and `uniformRs`:
   [#154](https://github.com/haskell/random/pull/154)
--- a/src/System/Random.hs
+++ b/src/System/Random.hs
@@ -42,6 +42,7 @@ module System.Random
   , uniformRs
   , uniformList
   , uniformListR
+  , shuffleList
   -- ** Bytes
   , uniformByteArray
   , uniformByteString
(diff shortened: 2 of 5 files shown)


Bodigrim (approved):
LGTM!

Shimuuar:
In general +1 but I have few comments. What are stability guarantees for the shuffle? Do we promise to change algorithm output only on major updates?

### Bias 

Shuffle is very slightly biased. For example let look at shuffle of `[a,b]`. List of indices (`uniformListM n gen`) will yield `2^N` ties out of `2^2N` possible options. Sort is stable so `[a,b]` will have probability of `1/2 + 2^{-N}` and `[b,a]` — `1/2 - 2^{-N}`.
Bias is negligible ~5e-20 in 64-bit case and small ~2e-10 in 32-bit case.

Analysis for n-element list is more complicated. Probability of having at least one collision in indices is subject to birthday paradox. So maybe it's possible to detect bias with 32-bit Ints. And yes Fisher-Yates shuffle is unbiased, has O(n) complexity but requires in-place mutation. It's awfully inconvenient to not having access to proper arrays


P.S. I'm not sure that `uniformList` warrants inclusion since we have `replicateM` but I'm not against it

lehins:
> Shuffle is very slightly biased. ..... It's awfully inconvenient to not having access to proper arrays

That's precisely why I didn't bother writing it using mutable arrays. I don't think this issue alone warrants a dependency on `primitive`.

> What are stability guarantees for the shuffle? Do we promise to change algorithm output only on major updates?

Yes. That would be the case. I would rather not change the underlying semantics on non-breaking version updates.

Shimuuar:
I'm on a fence about this. Arrays are after all very basic functionality. It's first time we run into algorithm that requires arrays. Will it be another one? Something that requires lookup tables, such as `mwc-random`'s normal.

Bodigrim:
If you like pain, you can use naked [`Data.Array.Byte`](https://hackage.haskell.org/package/base-4.18.0.0/docs/Data-Array-Byte.html), it should be enough to shuffle a list of indices Otherwise depending on `array` is less of a burden than `primitive`, because `array` is preinstalled as a boot package anyways.

lehins:
>  Otherwise depending on array is less of a burden 

I'd rather use my own redefinition of `data Array a = Array (Array# a)` then suffer through using `array` package :smile:

PR #118, opened 2021-09-05
Improve uniform ShortByteString (fixup)

lehins:
This PR attempts to fix a bug that sneaked in the #116 that affects Big Endian machines

I've tested it on 32/64bit Intel and 32bit armv7 machines, however both of these are little endian and I still have not figured out a way to test big endian code. @juhp, maybe you can help us out once more and run this PR on a s390x again? Thanks to your feedback in https://github.com/haskell/random/pull/116#issuecomment-911600622 I am pretty confident that this change will do the trick.

PR #116, opened 2021-08-31
Improve uniform `ShortByteString`

lehins:
random-1.2.0 contained a shortcut where architecture independent generation of bytes relied on `bytestring`'s builder functionality, which forced us to generate `ShortByteString` as pinned. This PR fixes that technical debt.

This is a non-breaking change.

PR #143, opened 2023-09-10
Add big-endian CI job

Bodigrim:
Following up on #116 and #118 saga, let's add a big-endian CI job, similarly to what `bytestring` and `text` do.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -244,3 +244,28 @@ jobs:
         name: Run build (arm64v8 linux)
         with:
           args: sh -c "cabal update && cabal test"
+
+  # Emulation on s390x platform is incredibly slow and memory demanding.
+  # It seems that any executable with GHC RTS takes at least 7-8 Gb of RAM, so we can
+  # run `cabal` or `ghc` on their own, but cannot run them both at the same time, striking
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yaml
```diff
@@ -244,3 +244,28 @@ jobs:
...
+        distro: ubuntu_rolling
+        githubToken: ${{ github.token }}
+        install: |
+          apt-get update -y
+          apt-get install -y ghc libghc-tasty-smallcheck-dev libghc-tasty-hunit-dev libghc-splitmix-dev curl
+        run: |
+          ghc --version
+          ghc --make -isrc:test-legacy -o legacy test-legacy/Legacy.hs
+          ./legacy
+          ghc --make -isrc:test -o spec test/Spec.hs
```
lehins, line 270: Since we don't build `random`, doesn't it mean that it will actually use the random version from the ghc packages store?

Which begs the question of why would we test a released version of `random` on CI?

lehins (commented):
Unless I am misunderstanding what is going on, I am not sure we want to have such an action.

.github/workflows/ci.yaml
Bodigrim: We do build `random`, just without using `cabal`.

Shimuuar (approved):

PR #153, opened 2023-10-29
Add ByteArray generation

### LINKED ISSUE #142: Swap genShortByteString for genByteArray
Since `bytestring-0.12` [`ShortByteString`](https://hackage.haskell.org/package/bytestring-0.12.0.2/docs/Data-ByteString-Short.html#t:ShortByteString) is just a newtype over [`ByteArray`](https://hackage.haskell.org/package/base-4.18.0.0/docs/Data-Array-Byte.html#t:ByteArray), a common data type shared by `base`, `text` and `primitive`. 

I think we can make `class Random` more useful if we swap

```diff
-genShortByteString :: Int -> g -> (ShortByteString, g)
+genByteArray Int -> g -> (ByteArray, g) 
```

This way we can shave off dependency on `bytestring` and do not impose it on our users. 

There does not seem to be many [users of `genShortByteString`](https://hackage-search.serokell.io/?q=genShortByteString) in the wild. Potentially we can provide a compatibility shim:

```haskell
genShortByteString :: Coercible a ByteArray => Int -> g -> (a, g)
genShortByteString = coerce genByteArray
```

lehins:
Fixes #142 and more.

Add compatibility with recently added `ByteArray` to base and fallback onto `data-array-byte` package for older ghc versions.

Besides this we also:

* Adds functionality for filling a `MutableByteArray`, instead of creating an immutable `ByteArray`. This can help avoid allocations in some situations when array has already been preallocated.
* Adds ability to generate pinned and unpinned `ByteArray`s.


lehins:
@Bodigrim Let me know what you think about this PR if and when you have some time.

Bodigrim:
@lehins overall looks great! Could you possibly organise code mentioning `ByteString` and `ShortByteString` together?

lehins:
> Could you possibly organise code mentioning ByteString and ShortByteString together?

@Bodigrim I am not sure I understand in which way you'd like it to be organized. All of the ByteArray/ShortByteString/ByteString functionality is together in the same part of the module. It probably doesn't look like that in the diff, because some of the `ShortByteString`/`ByteString` functionality hasn't changed and is not part of the diff.

PR #160, opened 2024-01-27
Introduce new `SplitGen` type class

### LINKED ISSUE #97: Make a better story for split
I think we should look to `Data.Bits` for inspiration here. Add a class

```haskell
class RandomGen g => SplittableRandomGen g where
  safeSplit :: g -> (g, g)
```

`safeSplit` is the same as `split`, except it should only be defined for generators that are actually splittable.

Question: can we improve efficiency or flexibility for any mutable generators by offering something like this?

```haskell
class StatefulGen g m => SplittableStatefulGen g m where
  splitStatefulGenM :: g -> m g
```

I imagine this might be able to avoid some unnecessary copying in some cases.

lehins:
Fixes #97 

I've been thinking for a while about this and I think the correct path forward is to introduce a separate class for splittable generators, which I called `SplitGen` and deprecate the old `split` function.

The benefit of this approach is that current user code does not break and gives a graceful path to upgrade to the new type class.

Other approaches were considered in #94 and discussed at length in https://github.com/idontgetoutmuch/random/issues/7

I highly doubt that we will provide a generic approach of using some hash function for implementing PRNG splitting. I would rather prefer such functionality be in a separate library that other packages that want to implement `SplitGen` can opt-in into using.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,6 @@
 # 1.3.0
 
+* Add `SplitGen` and `splitGen`
 * Add `shuffleList` and `shuffleListM`: [#140](https://github.com/haskell/random/pull/140)
 * Add `mkStdGen64`: [#155](https://github.com/haskell/random/pull/155)
 * Add `uniformListRM`, `uniformList`, `uniformListR`, `uniforms` and `uniformRs`:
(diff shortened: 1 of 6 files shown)


Shimuuar:
+1 

Following will cause compile failure for instances which don't define `split` and `SplitGen` instance. But I think it's fine
```haskell
  default split :: SplitGen g => g -> (g, g)
  split = splitGen
```

PR #165, opened 2024-07-29
Fix doctests

sol:
This addresses https://github.com/commercialhaskell/stackage/issues/7493.


src/System/Random/Internal.hs
```diff
@@ -956,7 +956,7 @@ class Uniform a where
   --
   -- There is a default implementation via 'Generic':
   --
-  -- >>> :set -XDeriveGeneric -XDeriveAnyClass
+  -- >>> :seti -XDeriveGeneric -XDeriveAnyClass
```
sol, line 959: @Bodigrim `:set` sets options for source files and `ghci` expressions, while `:seti` only affects `ghci` expressions.

1. With `:set` those three tests were failing, I think due to a reincarnation of https://gitlab.haskell.org/ghc/ghc/-/issues/20670.
1. For doctests you usually want `:seti` anyway.

sol:
@idontgetoutmuch I don't think the seed was fixed for those  examples before this PR.

lehins:
> Why do the generated random numbers change? I would have thought with a fixed seed we should always get the same random numbers i.e. why this: https://github.com/haskell/random/pull/165/files#diff-fe93ae678c398cfea72ff60186cc2cee27bd752aad99d8fd1ece059c912f0104L645

@idontgetoutmuch That is because that example uses global StdGen, which as comment below example points out, is usually not a very good idea:

https://github.com/haskell/random/blob/623cf518c6bc4c1de6a4d59b43c3ae37fcb22a35/src/System/Random.hs#L647-L650

src/System/Random.hs
```diff
@@ -354,7 +354,7 @@ class Random a where
   -- independently:
   --
   -- >>> fst $ randomR (('a', 5.0), ('z', 10.0)) $ mkStdGen 2021
-  -- ('t',6.240232662366563)
+  -- ('t',6.240232662366564)
```
lehins, line 357: @idontgetoutmuch I am more curious to see floating point number generation be affected by the change of how doctests are being executed. I suspect this is coming from some ghc optimizations, so I don't think we'll be able to do anything about it, but it could be worth investigating where exactly it is coming from.

.github/workflows/doctest.yml
```diff
@@ -0,0 +1,18 @@
...
+  doctest:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+
+      - run: cabal update
+
+      # NOTE: We can't use `cabal doctest` here, as `cabal doctest` uses
+      # `--build-depends=QuickCheck`, which results in a dependency cycle.
+      - run: cabal install doctest --ignore-project --overwrite-policy=always && cabal build && cabal repl --build-depends=unliftio --with-compiler=doctest --repl-options='-w -Wdefault'
```
lehins, line 18: @sol This approach will only run doctests for the latest ghc and latest cabal that is provided on the ubuntu image. Would it be possible, instead of adding a whole new workflow, to add this line to the cabal section of the GihubActions workflow here: https://github.com/haskell/random/blob/623cf518c6bc4c1de6a4d59b43c3ae37fcb22a35/.github/workflows/ci.yaml#L87

lehins (commented):
Thank you for fixing `doctest` for us @sol. I greatly appreciate it.
I could never figure out what was wrong with our current setup.
Could I ask you to change the CI in such a way that we run doctests for many ghc versions, instead of just one. This is quite important for making sure our examples also work the same for users of all supported ghc versions.

.github/workflows/doctest.yml
sol: > @sol This approach will only run doctests for the latest ghc and latest cabal that is provided on the ubuntu image.

Which is basically, with very short delays, the latest released version of ghc and cabal.

> Would it be possible, instead of adding a whole new workflow, to add this line to the cabal section of the GihubActions workflow

That makes sense 👍

I don't have the bandwidth to do this myself + I do think this PR is an improvement over the status quo. You may choose to merge this as-is, then somebody can improve on it in a follow up PR. Alternatively, anybody is welcome to take this PR as a template or starting point for a new PR.

I am happy to review any doctest related changes so please always feel free to @-mention me.  I just don't have the capacity to fix everybody's doctests myself.

The important points are:
- Don't use `:set`, use `:seti` instead.
- Don't use `doctest` as a library. Use `cabal repl` or `cabal doctest`. https://github.com/sol/doctest?tab=readme-ov-file#running-doctest-for-a-cabal-package

sol: Yet an other option, I granted "maintainer access" to my branch. So I think technically a maintainer could add new commits to my branch (or amend existing commits).

Shimuuar:
In fact we do fix seed for doctests. In module `System.Random.Stateful`:
```
-- $setup
-- >>> import Control.Monad.Primitive
-- >>> import qualified System.Random.MWC as MWC
-- >>> writeIORef theStdGen $ mkStdGen 2021
```

Maybe cabal doctest executes modules in different order?

.github/workflows/doctest.yml
lehins: No worries, @sol, I'll take it from here. You've been tremendously helpful with this PR already. So, thank you very much!

> I am happy to review any doctest related changes so please always feel free to @-mention me. 

Thank you very much, I'll definitely take you up on that offer! :slightly_smiling_face:

lehins:
> Maybe cabal doctest executes modules in different order?

I am sure that is the case. Which is also the reason why I always suggest avoid using the global generator whenever reproducability is important.

lehins:
> Are there any release plans, btw?

@juhp Yes, I am planning on making a release over Christmas.

In any case, I suspect stackage will be stuck on this for a while, since last time major version of `random` was released it took stackage almost a year to adopt it (all due tons of speculative upper bounds in downstream packages: https://github.com/commercialhaskell/stackage/issues/5474)

I can try making a patch release of `random-1.2.1.3` that drops usage of `doctest`, in order to unblock `doctes`. Would that be helpful?

juhp:
> I can try making a patch release of `random-1.2.1.3` that drops usage of `doctest`, in order to unblock `doctes`. Would that be helpful?

Yes that sounds good if you can, thank you!

juhp:
For now I try to disable the random testsuite in Stackage Nightly as a workaround hopefully

lehins:
> Yes that sounds good if you can, thank you!

@juhp [`random-1.2.1.3`](https://hackage.haskell.org/package/random-1.2.1.3) that does not depend on `doctest` is up on Hackage.

PR #167, opened 2024-10-26
Upgrade CI dependencies

lehins:
This PR:

1. Updates versions of different GithubActions tools
2. Adds ghc-8.10 to CI and upgrades others
3. Removes doctest CI run introduced in #165 in favor of the same doctest run that is executed for all cabal builds in the matrix.
4. Adds `Uniform` instance for `Maybe` and `Either` types. There is already a precedence for such instances, since any type that depends on `Maybe` or `Either` can already be derived generically. In fact ghc-8.10 is the only one that isn't capable of doing that, which resulted in a doctest failure, for [a test that depends on `Uniform (Maybe Bool)`](https://hackage.haskell.org/package/random-1.2.1.2/docs/System-Random-Stateful.html#v:uniformM).


src/System/Random/Internal.hs
```diff
@@ -1474,6 +1474,18 @@ instance UniformRange Bool where
   {-# INLINE uniformRM #-}
   isInRange = isInRangeOrd
 
+instance Uniform a => Uniform (Maybe a) where
```
Shimuuar, line 1477: I don't think Maybe and Either should get such instance. It doesn't result in uniform distribution. For example for `Maybe Word8`, `Nothing` will be produced with probability 1/2 and `Just i` with 1/2·256. Very far from uniform!

On other hand it we manage somehow to produce uniform distribution, it unlikely that it would be practical. For `Maybe Int64` probability of getting Nothing will be about 2^{-64}. 

In general uniform distribution of sums (unlike products) make little sense in general case. There're

lehins: See my comment in the description. Unfortunately we already do that whenever we derive instance generically, see doctest: https://hackage.haskell.org/package/random-1.2.1.2/docs/System-Random-Stateful.html#v:uniformM

lehins: @Shimuuar FYI, I do agree with you. However, because of this generic deriving mechanism for sum types we have accepted semantics as in the instances for Maybe and Either, that I've added in this PR. In my opinion we have to go one of the two routes:
1. Either remove mechanism for generic deriving for sum types
2. Or, accept that this is the way that Uniform instances should behave for sum types.

Shimuuar: But generic deriving at very least produce correct instances by counting all possible states using `Finite` type class. So at least it will produce correct distribution.

I still think these instances are more trouble that worth. But if we're going to add them let at very least make them produce correct distribution

lehins: To expand a little bit further. It is impossible to treat `Nothing` as a single case in all possible cases of `Maybe a`, because it is impossible to know how many cases `a` has. This problem essentially applies to all sum types with type arguments. This means we either reduce usability of `random` by treating such instances as unlawful. Or we could accept this and have `random` package be more versatile. I am leaning towards avoiding "perfection is the enemy of progress" and accepting such instances as legal.

lehins: Actually.  It looks like I was totally wrong. Those are not the same semantics that I have implemented. I will derive instances for `Maybe` and `Either` using generic mechanism instead

lehins: > But generic deriving at very least produce correct instances by counting all possible states using Finite type class. So at least it will produce correct distribution.

You are totally right. I completely forgot how the generic mechanism works with the help of `Finite`.

> I still think these instances are more trouble that worth. But if we're going to add them let at very least make them produce correct distribution

I've updated those instances. SO, now the distribution will be correct:

```haskell
ghci> runStateGenT_ (mkStdGen 2024) (uniformListM 20) :: IO [Maybe Word8]
[Just 76,Just 152,Just 79,Just 50,Just 186,Just 200,Just 115,Just 191,Just 159,Just 243,Just 170,Just 242,Just 98,Just 227,Just 63,Just 231,Just 124,Just 192,Just 174,Just 31]
ghci> filter isNothing <$> runStateGenT_ (mkStdGen 2024) (uniformListM 20) :: IO [Maybe Word8]
[]
ghci> filter isNothing <$> runStateGenT_ (mkStdGen 2024) (uniformListM 100) :: IO [Maybe Word8]
[]
ghci> filter isNothing <$> runStateGenT_ (mkStdGen 2024) (uniformListM 200) :: IO [Maybe Word8]
[Nothing]
```

lehins: @Shimuuar thank you for calling me out on this!

PR #172, opened 2024-12-23
Investigate floating generation

### LINKED ISSUE #53: incorrect distribution of randomR for floating-point numbers

### LINKED ISSUE #166: make `uniformRM (0, 1)` for floats equivalent to `uniformDouble01M` / `uniformFloat01M`
`uniformRM (l, h)` for `Float` and `Double` is defined by drawing `x` from [0, 1] using `uniformDouble01M` / `uniformFloat01M` and then returning `x * l + (1 - x) * h`.

One might expect that `uniformRM (0, 1)` may produce the same values as `uniformDouble01M` / `uniformFloat01M`. However, this is not the case because of rounding errors when calculating `1 - x`.

For visualization of the rounding problem try:
```hs
mapM_ print [ (x == x', x', x)
            | let m = fromIntegral (maxBound :: Word64) :: Double
            , i <- [0, 2 ^ 8 .. 2 ^ 14]
            , let x = fromIntegral i / m  -- resembles `uniformDouble01M`
            , let x' = 1 - (1 - x)
            ]
```
Note that the `x'` values have a lot of repetitions while each `x` is different.

An easy fix would be changing the calculation to `(1 - x) * l + x * h`. However, then the possible results of `uniformRM (-1, 0)` are reduced instead.

A possible solution could be to first (mathematically) transform the two calculations as follows:

1. `x * l + (1 - x) * h  =  h + x * (l - h)`
2. `(1 - x) * l + x * h  =  l + x * (h - l)`

Then use calculation 2 when `l` is closer to zero than `h`, and calculation 2 otherwise. This may be implemented as follows:

```hs
instance UniformRange Float where  -- analogously for `Double`
  uniformRM (l, h)
    | l == h                       = const $ return l
    | isInfinite l || isInfinite h = const $ return $! h + l
      -- Optimisation exploiting absorption:
      --   (-Infinity) + (anything but +Infinity) = -Infinity
      --   (anything but -Infinity) + (+Infinity) = +Infinity
      --                (-Infinity) + (+Infinity) = NaN
    | let f = if abs l < abs h then (let d = h - l in \ x -> l + x * d)
                               else (let d = l - h in \ x -> h + x * d)
      = fmap f . uniformFloat01M
  {-# INLINE uniformRM #-}
```
(This also drops the binding for the second parameter of `uniformRM` in order to allow memoization of the bound preprocessing.)

This approach also slightly reduces the documented [floating point number caveats](https://hackage.haskell.org/package/random-1.2.1.2/docs/System-Random-Stateful.html#g:fpcaveats), because this implementation guarantees that the bound that is closer to 0 is not exceeded.

Note: Obviously, this would break backwards compatibility as it changes the result of `uniformRM` for a given state of the RNG.

lehins:
This PR applies suggestion from and resolves #166 

It also fixes #53

It also makes the result of `uniformRM` on floating point numbers strict, since there is a huge performance penalty for keeping the result lazy


src/System/Random/Internal.hs
```diff
@@ -1438,16 +1442,46 @@ instance UniformRange Float where
...
+  -- ^ High
+  -> w
+  -- ^ Uniformly distributed unsigned integral value that will be used for scaling
+  -> a
+scaleFloating l h w =
+  if isInfinite diff
+    then let !x = fromIntegral w / m
+             !y = x * l + (1 - x) * h
+          in max (min y (max l h)) (min l h)
+    else let !x = fromIntegral (clearBit w 31) / m
```
Shimuuar, line 1476: Oh! 31 is for for Float, Double will need 63. We want to twiddle highest bit.

It seems that function doesn't generalize well. Maybe it's better to just create two monomorphic versions for Float and Double? It would be easier on compiler since it won't need to generate specializations at call sites

lehins: > Oh! 31 is for for Float, Double will need 63. We want to twiddle highest bit.

Nice catch. It is easy to solve generally with `finiteBitSize`

lehins: See updated version.

Shimuuar: LGTM!

src/System/Random/Stateful.hs
```diff
@@ -786,80 +787,97 @@ applyTGen f (TGenM tvar) = do
...
+--         @
 --
---     >>> :{
---     let (a, b, x) = (-1.9087862, -1.908786, 0.4228573)
---         result = x * a + (1 - x) * b :: Float
---     in (result, result < min a b)
---     :}
---     (-1.9087863,True)
+--         Because of this clever technique the result does not need clamping, since
+--         scaled values are guaranteed to stay within the specified range.
```
Flupp, line 839: Maybe also note that this solution, in contrast to `x * l + (1 - x) * h`, strives for losing less randomness to rounding by avoiding calculation of `1 - x` and similar.

AFAICT, the current solution works equally well if `l < h` as well as `l > h` (always returning numbers between `l` and `h`). Not sure if this is worth mentioning.

Flupp (commented):
Just some nits I noticed while looking at the current solution.

src/System/Random/Stateful.hs
```diff
@@ -786,80 +787,97 @@ applyTGen f (TGenM tvar) = do
...
 --
---     >>> let (a, b, x) = (nan, 1, 0.5) in x * a + (1 - x) * b
+--     >>> uniformRM (nan, 1) g
 --     NaN
---     >>> let (a, b, x) = (-1, nan, 0.5) in x * a + (1 - x) * b
+--     >>> uniformRM (-1, nan) g
 --     NaN
 --
--- *   If \(a\) is @-Infinity@ and \(b\) is @Infinity@, the result is @NaN@.
+-- *   If \(l\) and \(h\) are both @__Infinity__@ with opposing signes, then the result is @__NaN__@.
```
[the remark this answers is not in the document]
lehins: Thank you. Done

src/System/Random/Stateful.hs
```diff
@@ -786,80 +787,97 @@ applyTGen f (TGenM tvar) = do
...
---     Due to rounding errors, floating point operations are neither
---     associative nor distributive the way the corresponding operations on
---     real numbers are. Additionally, floating point numbers admit special
---     values @NaN@ as well as negative and positive infinity.
+--         Clamping is necessary, because otherwise it would be possible to run into a
+--         degenerate case when a scaled value is outside the specified range due to
+--         rounding errors.
 --
--- For pathological values, step 2 can yield surprising results.
+--     3.  Whenever @__h - l__@ does not overflow, we use this common formula ofor scaling: @__ l + (h - l) * x__@.
```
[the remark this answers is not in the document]
lehins: Done. Thank you.

src/System/Random/Stateful.hs
```diff
@@ -786,80 +787,97 @@ applyTGen f (TGenM tvar) = do
...
+-- procedure to generate a random value in a range for @uniformRM (l, h) g@:
+--
+-- * If @__l == h__@, return: @__l__@.
+-- * If @__`isInfinite` l == True__@ or @__`isInfinite` h == True__@, return: @__l + h__@
+-- * Otherwise:
+--
+--     1.  Generate an unsigned integral of matching width @__w__@ uniformly.
 --
--- If \(a = b\), return \(a\). Otherwise:
+--     2.  Check whether @__h - l__@ overflows to infinity and if does then convert @__w__@ to a floating
```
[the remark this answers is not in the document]
lehins: Done

src/System/Random/Internal.hs
```diff
@@ -1434,16 +1438,48 @@ instance UniformRange Float where
...
       return $! h + l
     | otherwise = do
-      x <- uniformFloat01M g
-      return $ x * l + (1 - x) * h
+      w32 <- uniformWord32 g
+      pure $! scaleFloating l h w32
   {-# INLINE uniformRM #-}
   isInRange = isInRangeOrd
 
+-- | This is the function that is used to scale a floating point values from random word range to
```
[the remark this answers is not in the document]
lehins: Done. Thanks

src/System/Random/Stateful.hs
lehins: > Maybe also note that this solution,

Done

> Not sure if this is worth mentioning.

I don't think it is necessary. It was only mentioned before, because those instances for floating point values were violating the laws imposed on `uniformRM`:

> Generates a value uniformly distributed over the provided range, which is interpreted as inclusive in the lower and upper bound.

PR #180, opened 2025-01-23
Document that `Data.Ix.inRange` is not a suitable default for `isInRange`

### LINKED ISSUE #178: Document differences between `Ix`'s `inRange` and `UniformRange`'s `isInRange`
When upgrading a package of mine to support `random-1.3.0`, I needed to add an implementation of [`isInRange`](https://hackage.haskell.org/package/random-1.3.0/docs/System-Random-Stateful.html#v:isInRange) to a `UniformRange` instance in order to make it typecheck. There already existed an `Ix` instance for the same type, and the `Ix` class offers a very similar-looking [`inRange`](https://hackage.haskell.org/package/base-4.21.0.0/docs/Data-Ix.html#v:inRange) method. As such, I was tempted to define `isInRange` in the `UniformRange` instance like so:

```hs
instance UniformRange T where
  ...
  isInRange = inRange
```

I later discovered that this is incorrect! This is because `UniformRange`'s `isInRange` requires the ranges to be symmetric, i.e.,

```hs
isInRange (lo, hi) x == isInRange (hi, lo) x
```

But this law does not necessarily hold for `Ix`'s `inRange`. This can be seen in the `UniformRange` and `Ix` instances for `Int`:

```
λ> isInRange (1, 0 :: Int) 0 -- UniformRange
True
λ> inRange (1, 0 :: Int) 0 -- Ix
False
```

There might be other differences between `isInRange` and `inRange` that I am not aware of, but this was one of the more prominent ones that I discovered. I worry that others may fall into the same pitfall. Would you be willing to document these differences in the Haddocks for `isInRange` to make them more obvious?

RyanGlScott:
Fixes #178.

### DIFF
--- a/src/System/Random/Internal.hs
+++ b/src/System/Random/Internal.hs
@@ -950,7 +950,12 @@ class UniformRange a where
   --
   -- There is a default implementation of 'isInRange' via 'Generic'. Other helper function
   -- that can be used for implementing this function are `isInRangeOrd` and
-  -- `isInRangeEnum`
+  -- `isInRangeEnum`.
+  --
+  -- Note that the @isRange@ method from @Data.Ix@ is /not/ a suitable default
+  -- implementation of 'isInRange'. Unlike 'isInRange', @isRange@ is not
+  -- required to be symmetric, and many @isRange@ implementations are not
+  -- symmetric in practice.
   --
   -- @since 1.3.0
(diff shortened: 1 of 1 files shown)


lehins (approved):
Thank you! This is a very nice addition to the documentation of `isInRange` function!

PR #192, opened 2025-09-05
Change to make it compile with MicroHs

augustss:
MicroHs does not have associated type ~synonyms~ families nor generics.
So there's a lot of extra CPP to work around that.


augustss:
You can close the PR.  You have been more polite about not wanting your package to be portable than some other package maintainers.
I have a fork that works.  I'll just start uploading my forked packages to hackage and add a mechanism to have `mcabal` redirect non-portable packages.
Thanks for the quick response.

PR #196, opened 2026-06-12
Addition of `MonadReader` like interface for `StatefulGen`

lehins:


Shimuuar:
That's great!

I played with code a bit. Few observations. `MonadStatefulGen` does not need `env` parameter, it's only used for default definition of functions in terms of MonadReader. And it could be dropped at modest cost of not having default implementation of `localGenM`:

```haskell
class (R.StatefulGen g m) => MonadStatefulGen g m | m -> g where
  askGen :: m g
  default askGen :: (MonadReader env m, HasGenEnv env g) => m g
  askGen = ...
  localGenM :: (g -> m g) -> m a -> m a
  default localGenM :: (MonadReader env m, HasGenEnv env g) => (g -> m g) -> m a -> m a
  localGenM = ...
```

Also I don't think `m` needs to be instance of `MonadReader env`. Let have 
```haskell
newtype Rand a = Rand (State SplitMix a)
```

Such type could obviously be instance of `MonadStatefulGen StateGenM` but there's no good reason to be instance of `MonadReader StateGenM` as well. 


#### DerivingVia

I have unreasonable fondness of DerivingVia so I checked whether it possible to use it too define instances. It turns out yet.
```haskell
newtype ViaReader env m a = ViaReader (m a)
  deriving newtype (Functor, Applicative, Monad)

deriving newtype instance MonadReader env m => MonadReader env (ViaReader env m)
deriving newtype instance R.StatefulGen g m => R.StatefulGen g (ViaReader env m)

instance (MonadReader env m, HasGenEnv env g, R.StatefulGen g m) =>
  MonadStatefulGen g (ViaReader env m)

deriving via ViaReader env (Lazy.StateT s m)
    instance ( R.StatefulGen g (Lazy.StateT s m)
             , MonadReader env m
             , HasGenEnv env g
             ) => MonadStatefulGen g (Lazy.StateT s m)

```
Also I didn't checked whether it actually works but it does compile.

